PR #5, opened 2013-11-22
Temporary fix for SUPNXP-9065: opening a file with ndrive that is not yet downloaded

jiakuan:
For this ticket https://jira.nuxeo.com/browse/SUPNXP-9065, I implemented a temporary fix for both Mac and Windows, and I manually tested on Mac OS X 10.8 and Windows 7 VM. Please see the screenshots attached.

screen shot 2013-11-22 at 12 45 23 pm
screen shot 2013-11-22 at 12 58 55 pm

I created a new branch, named 'SUPNXP-9065', based on the 'release-5.7.3' tag. (I tried to work with the latest master branch, but got some issues and just didn't bother to investigate that issues, so just worked with 5.7.3 source code).

Please note that, on win32 platform, when a separate ndrivew.exe process is launched by a 'edit' command, which is triggered by a user clicking on a custom URL 'nxdrive://...', the QApplication object is not initialized at all, so no system tray is initialised for this process. Therefore, it's not able to show message in system tray area as it does on Mac OS. I then implemented a temporary solution to launch a minimum app to show the message, and then quit automatically after about 10 seconds. Please let me know if there are any better solutions! If launching a temporary message app is an acceptable solution, then perhaps we can use a slightly different icon in system tray, like a 'nuxeo editor' icon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ MANIFEST
 .project
 .settings
 .pydevproject
+.idea
+**/.idea
--- a/nuxeo-drive-client/nxdrive/commandline.py
+++ b/nuxeo-drive-client/nxdrive/commandline.py
@@ -378,8 +378,34 @@ class CliHandler(object):
         return 0
 
     def edit(self, options):
-        self.controller.launch_file_editor(
-            options.server_url, options.item_id)
+        if sys.platform == 'win32':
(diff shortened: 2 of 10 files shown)


loopingz:
Old pull request on engine v2, thanks for your contribution

PR #15, opened 2014-10-29
Isolate cipher code and improve it

jowensla:
The proxy password was being encrypted and decrypted in the controller. Also it was using the token which can change easily.
I made a class utils.ControllerCipher and changed the secret to use the device_config.device_id which is more permanent. I added unit tests.


loopingz:
Pull request for engine v1 not applicable on engine v2, thanks for the contribution

PR #22, opened 2015-03-30
NXDRIVE-201: Request user attention via UI to handle update available.

IsaacYangSLA:


loopingz:
Please change the signal name to updateAvailable

PR #23, opened 2015-03-31
NXDRIVE-218: v1 code cleanup

ataillefer:


nuxeo-drive-client/nxdrive/tests/common.py
```diff
@@ -214,10 +213,12 @@ def get_all_states(self, session=None, get_pair_state=False):
         """Utility to quickly introspect the current known states"""
```
loopingz: Should be completely removed no ?

ataillefer: It's used in TestSynchronization to check the db states (so this test when reactivated should be a unit test), so I think we should refactor this function, or maybe there's already one on the framework to get the states?

PR #27, opened 2015-04-30
NXDRIVE-227: uninstallation might encounter missing registry, throws exc...

IsaacYangSLA:
...eption and does not remove nxdrive_home

### DIFF
--- a/nuxeo-drive-client/nxdrive/osi/windows.py
+++ b/nuxeo-drive-client/nxdrive/osi/windows.py
@@ -23,20 +23,10 @@ class WindowsIntegration(AbstractOSIntegration):
         try:
             key = _winreg.OpenKey(reg, path, 0, _winreg.KEY_ALL_ACCESS)
         except Exception, e:
-            # Sometimes the app is installed, but not run 
-            # Hence the path argument is invalid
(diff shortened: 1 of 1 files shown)


nuxeo-drive-client/nxdrive/osi/windows.py
```diff
@@ -20,9 +20,26 @@ def get_menu_key(self):
...
-        key = _winreg.OpenKey(reg, path, 0, _winreg.KEY_ALL_ACCESS)
-        _winreg.DeleteValue(key, value)
+        try:
+            key = _winreg.OpenKey(reg, path, 0, _winreg.KEY_ALL_ACCESS)
+        except Exception, e:
+            # Sometimes the app is installed, but not run 
+            # Hence the path argument is invalid
+            # It throws WindowsError 2: The system cannot find the file specified
+            # The uninstallation doesn't complete if exception is not caught
+            print e
```
loopingz: I suppose the print was used because log is not available while uninstall, do you think we can get rid of them ?

IsaacYangSLA:
Yes.  They are mainly for my debugging.  I will remove those print statements.

loopingz:
Thanks

PR #28, opened 2015-05-07
NXDRIVE-227: refactor caption so it can be customized

IsaacYangSLA:


nuxeo-drive-client/nxdrive/wui/dialog.py
```diff
@@ -559,7 +559,7 @@ def browse_folder(self, base_folder):
             local_folder_path = base_folder
             # TODO Might isolate to a specific api
             dir_path = QtGui.QFileDialog.getExistingDirectory(
-                caption='Select Nuxeo Drive folder location',
+                caption=self._get_caption_browse_folder(),
```
loopingz: The best way is to use the Translator service i think

loopingz:
I used the Translator instead, but thank you for spotting this issue

PR #36, opened 2015-07-23
SUPNXP-14086: Blank window is shown when property icon is clicked

tgopalakrishnan:


ataillefer:
OK thanks, tracked by https://jira.nuxeo.com/browse/NXDRIVE-394 and fixed by https://github.com/nuxeo/nuxeo-drive/commit/00efca54a21771f071ea853e35fbb90a92d26d60

PR #37, opened 2015-07-25
SUPNXP-14117: System tray window entry should not be created in task bar

arameshkumar:


loopingz:
Merge with a different ticket name as SUPNXP are private, we should coordinate so I can give you a ticket number to commit against.
https://github.com/nuxeo/nuxeo-drive/commit/a3073b97932a84c65c74ca3f64dde60f0c746012

Thanks for the contribution

PR #38, opened 2015-08-04
NXDRIVE-393: Display user full name in conflict management window

arameshkumar:


nuxeo-drive-client/nxdrive/wui/dialog.py
```diff
@@ -678,8 +678,8 @@ def __init__(self, application, page, title="Nuxeo Drive", api=None, token=None)
             self._api = api
         self._attachJsApi()
         self._frame.javaScriptWindowObjectCleared.connect(self._attachJsApi)
-        self.resize(400, 400)
-        self._view.resize(400, 400)
+        self.resize(550, 600)
```
loopingz: This is not related to NXDRIVE-393

PR #48, opened 2015-09-15
Add RPM specific actions to README.md

DirkHoffmann:
Unfortunately I could not get it fully running yet, as my server does not use uid/pwd credentials, but certificate/SSO.
However the basic installations commands are not difficult to translate. I hope I will be able to report successful running on Scientific Linux and Fedora soon.


ataillefer:
Thanks!

PR #52, opened 2015-09-29
NXDRIVE-438: Update set proxy to return string

rsoumyassdi:
@loopingz 
Bonjour,
I have updated the code to return a value from ```set_proxy_settings``` under settings.py. And I have overriden the same method under ```CPOManager``` class to validate the proxy, which handles the result that the UI projects to the user. So inorder to return the value(Valid or Invalid) I had to change this part.

CPO side code changes have been already made. 
https://github.com/nuxeo/sharp-cpo-clients/pull/154

### DIFF
--- a/nuxeo-drive-client/nxdrive/manager.py
+++ b/nuxeo-drive-client/nxdrive/manager.py
@@ -763,6 +763,7 @@ class Manager(QtCore.QObject):
         proxy_settings.save(self._dao)
         self.refresh_proxies(proxy_settings)
         log.info("Proxy settings successfully updated: %r", proxy_settings)
+        return ""
 
     def refresh_proxies(self, proxy_settings=None, device_config=None):
         """Refresh current proxies with the given settings"""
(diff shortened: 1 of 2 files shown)


loopingz:
Hi, you miss a part to return at least a value on the default manager for this pull request. You can either add the validation part in Drive or make the manager return always a empty string.

Regards

rsoumyassdi:
@loopingz 
If i got your point, we can add a return under manager.py ```set_proxy_settings()``` like,
```python
 def set_proxy_settings(self, proxy_settings):
       proxy_settings.save(self._dao)
       self.refresh_proxies(proxy_settings)
       log.info("Proxy settings successfully updated: %r", proxy_settings)
+      return ""
```
So that it always returns a value.

loopingz:
That's my point, I'll add the proxy validation code later, so be careful with your override.

Regards

rsoumyassdi:
Thanks a lot!

loopingz:
Thanks to you !

PR #58, opened 2015-11-09
fix-NXDRIVE-465: When setting folder icon use read-only attribute

rajadg:


ataillefer:
Hello,

Can you please explain:
- What is the purpose of setting the Nuxeo Drive root folder as READONLY instead of SYSTEM? Is the fact that it is marked as SYSTEM an issue? In this case what would be the issue?
- What is th link with [NXDRIVE-465](https://jira.nuxeo.com/browse/NXDRIVE-465): "Ignore 'hidden' files on Windows"?

Thanks.

mconstantin:
This pull request is generated by SSDI, but since they are not currently available, I'll try to comment on this.
I think the reference to NXDRIVE-465 is wrong, it should be https://jira.nuxeo.com/browse/SUPNXP-15086.
And the change applies not only to Nuxeo Drive root folder but all the Drive-marked read-only folders.
Seems that either "system" and "read-only" attributes cause Windows to read the desktop.ini in order to apply any folder customizations, e.g. the red icon.
Moreover, the "read-only" does not prevent the folder from being modified (is ignored by Windows, for folders). I don't know if "system" attribute prevents folder modification.
In short, I don't know the reason for changing from "system" to "read-only" for Drive designated read-only folders; we have to wait for Raja to answer this.

BryanSchacht:
I think the reason originally was with CPO we maintain 2 top level folders

CPO---My Docs
       ---Other’s Docs

If you put files in the folder CPO (above My Docs/Other’s Docs) they never will be synced at all.  So the thinking was to control write to that folder (or to any folder a user doesn’t have r/w permission) so they won’t add files that never sync (causes calls to support).

You are correct it can be overridden but not sure if there might still be a benefit to deny writes to users for some folders to avoid confusion.  If it doesn’t work or causes other problems I don’t want to add the code either but if it worked it might be ok.

Bryan

mconstantin:
I tested the original issue (DS menu showing conflict for 'Other Docs' (with Nuxeo commit for this issue), and I cannot reproduce it. So it seems that the previous commit fixes the issue in the https://jira.nuxeo.com/browse/SUPNXP-15086.

Using the SYSTEM attribute for read-only folders (i.e. DM shared folders with read-only permissions) prevents syncing content added to them in DS (because files/folders) with this SYSTEM attribute are ignored by DS. Additionally, DS also sets the SYSTEM attribute for folders created inside a folder with the SYSTEM attribute set, thus preventing their content from syncing, and so on.  
If we change the attribute used to mark "read-only" folders from SYSTEM to READONLY, then the content of those folder will be synced with DM, which is not desired since the folder is "read-only" (but cannot be enforced on the OS level, in Windows).
So I think this pull request is not necessary but I will let the author to comment otherwise.

walachintan:
As a conclusion from comments added on https://jira.nuxeo.com/browse/SUPNXP-15086 (that defect stays reopen). Any relation to this pull request needs to be derived by review. As of now, we feel this pull request does not resolve 15086.

mconstantin:
I take the previous comment back; this should be merged, see comment in https://jira.nuxeo.com/browse/SUPNXP-15086.

walachintan:
@ataillefer can you please commit this pull request. It has been pending for a while now.

ataillefer:
We first need to check the impact on a bare Drive and fully understand the problem.

tiry:
The PR does change the flag associated to folder when calling `set_folder_icon_win32`.
In Drive, `set_folder_icon_win32` is only called on Synchronization roots.

I do understand the fact that in the current code we set the roots as `SYSTEM`.
However, I do not understand the idea of making them `READ_ONLY`.

Does the DS code call `set_folder_icon_win32` on other folders that the roots ?

rajadg:
DS code calls the set_folder_icon_win32 on all read-only shared folders. This includes both "Other Docs" folder, sync roots and also sub folders of sync roots.

BryanSchacht:
On my system it is almost useless though. Properties say read only but Windows ignores that anyway so it seems not to accomplish the original purpose. I was hoping to ignore system files or hidden from upload sync as one goal. Second was to keep users from adding files to the root above My docs out adding to a read-only folder but that isn't working either. Only benefit is the red icon

loopingz:
The correct code should be to remove completely the attribute on the folder, you need to setup attribute on the .ini file, not sure readonly is needed. It was taken from already a code from SSDI in march, I should have verified more.

rajadg:
NXDRIVE-282 fixes this issue which is already merged. The fix from NXDRIVE-282 is almost same as this pull request so closing this one.

PR #66, opened 2016-01-07
SUPNXP-15494: sync up to server sometimes does not upload all files

rajadg:


loopingz:
If you look in the remote watcher the line just before your modification, the test was commented by myself one year ago to handle change of rights on readonly file, and probably lock.

BryanSchacht:
So what is the outcome?  Remi are you saying the test should be uncommented?  And if so will it have anything to do with multiple versions getting logged on the server or another situation where SIICA is getting multiple notifications on file uploads.  We'll probably enter a JIRA for it..

loopingz:
There is some modification on readonly and locks that has been made lately by Antoine, also the PR comes without a test to reproduce, so it is hard to merge as it is.

rsoumyassdi:
I tested DS Client with DM server, Below are the scenarios and the observations:
* With 5-10 Dirty file uploaded from DS Client > DS Client recieves only "documentCreated" event from DM in the GetChangeSummary response.(Expected)
* With 100-150 files uploaded from DS Client > For a few(4-5) files DS Client recieves both "documentModified" and "documentCreated" events in the same response of GetChangeSummary for that particular file.
Some of the files had their versions increased on DM.

To confirm the same, I have tested with Nuxeo Drive (2.1.113) with Base Nuxeo Server:
* With 5-10 Dirty file uploaded from Drive > Drive recieves only "documentCreated" event.
* With over 200 files uploaded from Drive > For 3-4 files Drive recieves both "documentModified" and "documentCreated"events in the same response in GetChangeSummary for that particular file.
Some of the files had their versions increased on DM. And a few files were duplicated for ex: file1__1.rtf

Considering the scenarios above, the issue appears to be similar in both DS client and Nuxeo Drive, only difference could be the count of files with which the issue was reproduced

loopingz:
This PR is now misaligned with the code and comes from a repository that has been delete, therefore i m closing it. Please reopen one if needed, sorry for the inconvenient

PR #69, opened 2016-01-22
Add git as needed package and update drive version

rverchere:
Git is needed to install nuxeo-drive on linux, using pip.


ataillefer:
Merged, thanks!

PR #74, opened 2016-03-09
PR/9203 use correct app version

mconstantin:
When base_automation_client communicates with the server, it gets its client_version (which constitutes the version part of the "User-Agent") from its creator, which is the Manager.   
The manager gets its version from the __init__.py of the root package.  
In DesktopSync, CPOManager inherits from the Nuxeo Manager, and has its own version in the __init__.py files under tits respective root package.

Sometimes the server is contacted before the CPOManager has a chance to set up the version.   
These calls uses Nuxeo version:
class CPOManager(Manager):
    def __init__(self, options):
        super(CPOManager, self).__init__(options)  <== server is contacted in the init process
        ...
        self.client_version = __version__
But subsequently all requests should use the CPOManager's version.

However, as observed in the field, some clients from the same build seems to send out different versions for the same server call.   
This change initializes the "manager" with a version provided at app launching (from commandline.py).

### DIFF
--- a/nuxeo-drive-client/nxdrive/commandline.py
+++ b/nuxeo-drive-client/nxdrive/commandline.py
@@ -474,6 +474,7 @@ class CliHandler(object):
 
     def get_manager(self, options):
         from nxdrive.manager import Manager
+        setattr(options, 'version', __version__)
         return Manager(options)
 
     def _get_application(self, options, console=False):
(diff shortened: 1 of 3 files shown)


loopingz:
I think it would be better to remove self.client_version variable and always use the get_version so you can override it

loopingz:
Ok perfect

PR #75, opened 2016-03-10
PR/9001 changes to blacklist

mconstantin:
If the server is too busy to respond and DS gets an error (e.g. 500, or 429/503 with the recent throttling mechanism using nginx), it give up after 3 tries in relative quick succession..  
A common thread is that customers see correct folders/files in DM after DS uploads but then a second DS client on same folder (or share the same folder) doesn’t get the update, from DM.  
- What happens if missing the event to update for the second DS client? 
- Is DM not including in changelist?
- Is DS not processing changelist due to server response timeout or error?
- Does DS recover by trying later or is the change lost forever?

This change increases the number of attempts to 15, using an "exponential back off" scheme (which extends the retry period to ~3d).

### DIFF
--- a/nuxeo-drive-client/nxdrive/engine/blacklist_queue.py
+++ b/nuxeo-drive-client/nxdrive/engine/blacklist_queue.py
@@ -2,15 +2,23 @@
 Created on 1 juil. 2015
 
 @author: Remi Cattiau
+
+Updated on Feb 17, 2016
+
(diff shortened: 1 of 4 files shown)


BoboTiG:
Closing as very old. Thank you for the PR, but these kind of changes has to be handled via our JIRA system. I would be happy to discuss about that there.

PR #76, opened 2016-03-10
Pr/8920 after server update files do not sync

jlumsla:
if (file) upload fails with an error detail of the type {…“error” : “Unable to find batch..”}, then clear the cookies and retry both getting the batch id and the upload. 

This occurs after a server update is completed.

The suspected reason is that the batch/upload call went to one DM instance and the batch/execute call is routed to another DM instance by the AWS ELB. 

An invalid AWSELB cookie can cause this problem. An AWSELB cookie can become invalid if the DM instance that corresponds to the AWSELB is no longer alive (i.e. terminated). If the DS client uses an invalid AWSELB cookie, then the load balancer will do random routing of the call to any DM instance. In this case, the old DM instances was terminated and new DM instances were created. The cookie value of all the DS clients became invalid. So the batch/upload and batch/execute calls were routed to different instances and thus file upload from DS failed.


loopingz:
Why do you have added 3 retries also ? To simplify your deploiement we can remove the cookie anytime it cannot find the batch on the server no ?

jlumsla:
3 retries may be overkill but was added for reliability in case of continued failures. Regarding your suggestion how does that work?

loopingz:
Based on your current work, just removing the loop of retries ( the retries should be handle by the engine not the client, it would add two layers of retry handling ), and keep your try: except: with the cookie cleaning in case of InvalidBatch

jlumsla:
I made changes based on review comments. Please review the changes. Thanks.

loopingz:
Merged through feature-NXDRIVE-556-reset-cookie-on-invalid-batch

PR #77, opened 2016-03-10
PR/8671-handle-remote-scan-exceptions

jlumsla:
Remote scan would restart from the beginning when IntegrityError exceptions are thrown. With many files this could happen many times. 
Handle the IntegrityError exceptions in insert_remote_state and update_remote_state methods.


loopingz:
We don't want swallow exceptions

PR #80, opened 2016-03-15
PR/SUPNXP-16258-corrupt-db

mconstantin:
Add some validation to states in the local db.
'local_state' for a row is corrupted, reason unknown.


loopingz:
As discussed with you in meeting, i think the case is too exceptional now to create a specific, it could probably happen also on path or any other field, and this fix won't prevent the failure in that case. I would probably choose to remap completely the database if such things happens.

PR #84, opened 2016-03-23
PR/SUPNXP-15494 test upload files with native local client

mconstantin:
Integration test for SUPNXP-15494 using a new Mac-native local client.


loopingz:
The override methods of normal LocalClient takes argument as absolute path, and the normal client takes path as relative, it then breaks the logic of the LocalClient

loopingz:
No more updates on that PR so closing it

PR #91, opened 2016-04-18
SUPNXP-15494: mac sync up to server does not upload all files

rajadg:


nuxeo-drive-client/nxdrive/engine/watcher/local_watcher.py
```diff
@@ -121,7 +121,7 @@ def _execute(self):
                     self._end_action()
                 while (not self._watchdog_queue.empty()):
                     # Dont retest if already local scan
-                    if not trigger_local_scan and self._watchdog_queue.qsize() > self._windows_queue_threshold:
+                    if AbstractOSIntegration.is_windows() and not trigger_local_scan and self._watchdog_queue.qsize() > self._windows_queue_threshold:
```
rajadg: If a local scan is triggered in Mac many files are not uploaded. The local scan is meant for windows, but it is not giving good results for Mac. So avoiding local scan in Mac in case of copying more than 50 files is better.

BoboTiG:
Closing: it *may* be outdated. Open a new ticket if needed :)

PR #93, opened 2016-05-31
NXDRIVE-50: Update setup.py to handle pip register/upload/bdist_wheel

loopingz:


setup.py
```diff
@@ -7,7 +7,10 @@
 import sys
 from datetime import datetime
 
-import nx_esky
+try:
+    import nx_esky
+except Exception as e:
+    print e
```
ataillefer: What error do you get at esky import?

PR #98, opened 2016-07-27
Fix nxdrive 622 conflicts window shows blank screen for long time

mkeshava:


loopingz:
A different solution was picked to avoid any blank screen : showing the conflicts with user id and resolving the username asynchronously

PR #99, opened 2016-07-27
Up/master

mconstantin:
Refactoring breaks down the if...elif...else... in the _is_ignored_ method into separate "filters", making it modular.
As a result, no need to override the method in _CPOLocalClient_ to ignore something else.
Just add another method **anywhere**  in the source code (example for an instance method):
```Python
from nxdrive.client.local_client import register
...
class CPOLocalClient(LocalClient):
    # Limit to 5GB files
    SIZE_LIMIT = 5 * 1024 * 1024 * 1024

    @register(max_size=SIZE_LIMIT)
    def ignore_by_size(self, parent, name, **kwargs):
        max_size = kwargs['max_size']
        path = self._abspath(os.path.join(parent, name))
        stat = os.stat(path)
        return stat.st_size > max_size
    ...
    # other methods here
```
For a module level method, just pass None for the first argument (assuming the local client is not needed, e.g. for calculating the absolute path).  
Use _apply_to_parent_ decorator argument to allow a filter to apply to file's parent as well.
This way, files in an "ignored" folder will be ignored as well (new); this could be applied to the "prefix" rule (because folders starting with '.' are hidden on Mac) and to the "hidden attribute" rule (Windows).  
Note that this will also ignore folders starting with '~$' or the word 'Icon', etc. To fix this may create another "ignore filter" function and apply only the '.' prefix to it (and also the _apply_to_parent=True_).


mconstantin:
This is related to https://jira.nuxeo.com/browse/SUPNXP-14915

efge:
You should create a NXDRIVE JIRA issue describing the goal of the refactoring and include your detailed explanation in the description of the NXDRIVE. Then reference that NXDRIVE in the commit message.

Also create another NXDRIVE JIRA issue for the mac client and reference it in that commit message.

What does CSPII-11513 refer to in the other commit message?

mconstantin:
If I try to create a NXDRIVE JIRA, it shows **only** the "Sharp Nuxeo Connect" project available, not "Nuxeo Drive", which I think will create a SUPNXP-xxxxx JIRA.

efge:
Ah yes. So please create SUPNXP issues and I'll move them to NXDRIVE so we can have public tickets for your changes.

PR #104, opened 2016-09-15
SUPNXP-17993: fix error installing the segmentation fault handler…

mconstantin:
… (as seen in the Jenkins log for OS X build)


loopingz:
Fixed with similar approach

PR #106, opened 2016-09-16
CSPII-11999: Add exception type to exception handler in unbind

jlumsla:
Please review and merge.


loopingz:
Already added in october

PR #111, opened 2016-10-26
NXDRIVE-637: Drive local changes are not uploaded when resolving with local on conflict

arameshkumar:
I have modified the base nuxeo code to resolve this issue.


loopingz:
The code was already in the branch fix-NXDRIVE-637-resolve-local-fix

PR #113, opened 2016-10-29
NXDDRIVE-634: Update file move condition

RavinderSharp:
1. Update the move detect condition.
2. When user has "Read Only" permission on source document, the undelete operation will throw exception. Catch exception and upload as new file.


loopingz:
Part of this PR has been integrated with some changes, and I fixed the commit message

PR #116, opened 2016-11-22
fix-CSPII-12337: reset cookie jar for every use of the base automatio…

mconstantin:
…n client. Increase # of processors.

### DIFF
--- a/nuxeo-drive-client/nxdrive/client/base_automation_client.py
+++ b/nuxeo-drive-client/nxdrive/client/base_automation_client.py
@@ -10,6 +10,7 @@ import os
 import hashlib
 import tempfile
 from urllib import urlencode
+from cookielib import CookieJar
 from poster.streaminghttp import get_handlers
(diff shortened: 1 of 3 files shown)


loopingz:
The cookie is used for servers that need sticky session, for those who don't need sticky session then the load balancer simply ignored the cookies.
The sticky session is need with non-shareable transient store.

mconstantin:
The cookie is still there and being used when needed, e.g. for file uploads consisting of 2 requests (to /batch/upload followed by /batch/execute).
But it is also being preserved across all remote clients making file uploads (or other requests), completely unrelated. Basically all remote clients for the lifetime of the app, share the same cookies with the first used remote client; that is because they all share the same (engine's) cookie jar, which is also never cleared.

loopingz:
Having a shared instance of a cookie is exactly the same model as any browser, in your case it looks like you try to remove stickyness on your LoadBalancer by multiplying cookies on the client side.

PR #118, opened 2016-12-05
NXDRIVE-678: Keep the parent folder when unselecting the only child f…

mkeshava:
…older from the Select Sync window


mkeshava:
Please review and merge the PR.

loopingz:
The behavior from Drive is completely normal, Nuxeo Drive list files and folders compare to DS which only list folders i think

Regards

loopingz:
After discussing with Drive PO, we think it is a good PR sorry about that

PR #119, opened 2016-12-14
Feature nxdrive 679 support pac script for proxy settings

rajadg:


nuxeo-drive-client/nxdrive/manager.py
```diff
@@ -6,6 +6,10 @@
 import sys
 import logging
 from urlparse import urlparse
+import urllib2
+import js2py.pyjs
```
rajadg: Need to import js2py.pyjs explicitly so that build script will include this module while packaging.

loopingz: You could also use the setup.py includes array, it would be better to avoid later cleanup for python p8p. What do you think ?

rajadg: Initially I tried` includes ` array option and still the build did not load pypac library properly. This option `import js2py.pyjs` worked well so I did not pursue the `includes` array option. I will try again.

rajadg: @loopingz 
I added `js2py.pyjs` to the includes array and tried building nuxeo-drive. Unfortunately the library.zip file in the generated build does not contain the js2py.pyjs and its related modules.
**Reason**: I suspect the includes option itself might have a bug. When I moved `includes` options to `bdist_esky` instead of `build_exe`, with some more changes the build succeeded. It would be best if I fix this particular import issue as a separate bug.
I have created another JIRA https://jira.nuxeo.com/browse/SUPNXP-18699 that consolidates some 3 issues related to the build errors in nuxeo-drive. I am waiting for a NXDRIVE ticket for this issue.
Please check the JIRA and let me know your thoughts.

rajadg:
Removed some extra import statements

PR #120, opened 2016-12-21
NXDRIVE-683: Improve build script

rajadg:


loopingz:
Merged with a small update : https://github.com/nuxeo/nuxeo-drive/commit/24fd283ac6a3fe4bae09b32870900862df10547f
Let me know if you think it should not be this way

PR #123, opened 2017-01-04
NXDRIVE-677:fix synchronization if folder path is greater than 260 ch…

mkeshava:
…aracters on different platforms

Please review and merge the changes.


loopingz:
Merged thank you for the contribution

PR #128, opened 2017-01-16
Remove use of deprecated and insecure os.tempnam()

BoboTiG:
[tempnam](https://docs.python.org/2/library/os.html#os.tempnam) is deprecated and insecure by design. [mkstemp](https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp) seems the best alternative.


ataillefer:
Might want to reference a JIRA issue here pointing to the Python doc explaining why it is deprecated.
Otherwise :+1:

PR #130, opened 2017-01-16
NXDRIVE-698: rename UnitTest assertEquals() to assertEqual()

BoboTiG:
It is a lot of changes, but now we have something uniform. no more mix of `assertEqual()` and `assertEquals()`.

I checked and there is no other deprecated aliases in use.


ataillefer:
:+1:

PR #134, opened 2017-01-19
[WIP] NXDRIVE-711: fix system tray menu behavior

BoboTiG:
[WIP, do **not** merge]

Several points about this PR:

- Reordered imports, and use `from PyQt4.xxx import yyy` to facilitate future transition to Qt5.
- Removed several too broad `Exception` as they are catched at a higher level, and the code is more readable now
- Several modifications to be more PEP8 compliant
- Several small refactors: the menu seems more reactive
- The `replace` method has been greatly simplfied. The menu acts more like it should: its position is based on the icon position, not the cursor one.
- Add GUI tests for the first time (still need work)

I tried on GNU/Linux and Windows and changes don't break something.
Need tests on Mac.

### DIFF
--- a/nuxeo-drive-client/nxdrive/tests/test_gui_systray.py
+++ b/nuxeo-drive-client/nxdrive/tests/test_gui_systray.py
@@ -1,0 +1,97 @@
+# coding: utf-8
+
+from tempfile import mkdtemp
+from unittest import SkipTest, TestCase
+
+from PyQt4.QtCore import Qt, QPoint
+from PyQt4.QtTest import QTest
+from mock import Mock
+
+from nxdrive.logging_config import configure, get_logger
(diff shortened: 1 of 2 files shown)


loopingz:
The Exception in the slot on the API called is to avoid blocking WebKit when an exception occured, they are not catched to my knowledge anywhere else

BoboTiG:
If I am right, they are catched into the parent class `WebDriveApi`, aren't they? Into the file `nuxeo-drive-client/nxdrive/wui/dialog.py `.

ataillefer:
Yes but not all of them, typically `suspend`, `resume`, `get_syncing_items`. Also, I don't know if we can guarantee that `self._dialog.close()` will never throw an exception.
Otherwise sounds fine.

PR #135, opened 2017-01-26
Feature nxdrive 530 use random annotation

loopingz:


nuxeo-drive-client/nxdrive/tests/common_unit_test.py
```diff
@@ -60,19 +60,63 @@
...
+            for i in range(0, self._repeat):
+                log.debug('Repeating test %s %d/%d', func.func_name, i+1, self._repeat)
+                try:
+                    res = func(*args, **kwargs)
+                    # In RELAX mode if the test success one we dont fail
+                    if self._mode == 'RELAX':
+                        return res
+                except Exception as e:
+                    # In STRICT mode we propagate Exception
+                    if self._mode == 'STRICT':
```
jcarsique: if not STRICT, would it be interesting to collect the exception?

PR #137, opened 2017-02-03
NXDRIVE-724: Configuring manual proxy should not set global proxy con…

rajadg:
…figuration during validation


BoboTiG:
:+1:

ataillefer:
:+1:

PR #142, opened 2017-02-27
NXDRIVE-742: Remove obsolete LocalClient classes

BoboTiG:
With NXDRIVE-726, new errors apparaired. They are now fixed.


loopingz:
Why removing the specific test client that use the API of the OS ?

PR #145, opened 2017-03-06
NXDRIVE-742: Add native LocalClient tests to simulate Drive operations

BoboTiG:
Also:
    - Optimize imports
    - Several PEP8 fixes

Real changes starts at ligne 245 in test_local_client.py. I renamed the class `TestLocalClient` to `StubLocalClient` and added two classes that inherit of it. Tests are declared into that class.

Then to test OS specific operations (like Explorer or OSX File Manager), the class `TestLocalClientSimulation` overrides few test methods to explicitely fail when needed. And the class `TestLocalClientNative` uses the native/normal `LocalClient` as Drive would do to test Drive operations.

Finally GNU/Linux does not need to test specific behaviors because it uses native LocalClient by default.

### DIFF
--- a/nuxeo-drive-client/tests/mac_local_client.py
+++ b/nuxeo-drive-client/tests/mac_local_client.py
@@ -1,69 +1,76 @@
-from nxdrive.client.local_client import LocalClient
-'''
-Intent of this file is to use OSX File Manager to make FS operation
-'''
+"""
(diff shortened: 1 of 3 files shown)


BoboTiG:
:+1:

ataillefer:
:+1:

PR #154, opened 2017-03-10
NXDRIVE-758: Add integrity checks and several download tries

BoboTiG:
It should help to prevent failure at setup step. I added a MD5 checksum and 5 download tries.

### DIFF
--- a/tools/posix/deploy_jenkins_slave.sh
+++ b/tools/posix/deploy_jenkins_slave.sh
@@ -50,6 +50,20 @@ check_import() {
     echo "OK."
 }
 
+check_sum() {
+    # Calculate the MD5 sum of the file to check its integrity.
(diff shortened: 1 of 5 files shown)


tools/posix/deploy_jenkins_slave.sh
```diff
@@ -100,10 +114,27 @@ download() {
...
+            if check_sum "${output}"; then
+                return
+            fi
+            rm -rf "${output}"
+        fi
+        echo ">>> [$try/5] Downloading $url"
+        echo "                   to $output"
+        set +e
+        curl --silent -L "$url" -o "$output"
+        set -e
```
efge: You should merge these 3 lines and avoid the explicit `set` statements (which don't belong in the middle of a script) using just `curl --silent -L "$url" -o "$output" || true`

PR #159, opened 2017-03-15
NXDRIVE-766: Skip temporary test_conflicts.py failing due to NXP-21854

BoboTiG:
It is temporary for we can work on other issues while the fix will be ready.


loopingz:
👍

PR #179, opened 2017-03-27
NXDRIVE-747: Insert the current path at the first place when building…

BoboTiG:
… executable

This was the cause of auto-update endless loop. If Nuxeo Drive was already
installed via pip on the system, the package builder was shipping that
Nuxeo Drive instead of the one from fresh sources.


efge:
👍

PR #191, opened 2017-03-31
NXDRIVE-571: Remove use of "triggerRemoteJob" for "build"

BoboTiG:
Much better now. If the triggered job fails, the current job will too.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -60,37 +60,35 @@ change_version() {
 create_beta() {
     local version
     local drive_version
+    local git_opt
 
     version="$(grep -Eo "[0-9]+.[0-9]+" nuxeo-drive-client/nxdrive/__init__.py | tr '\n' '\0')"
     drive_version="${version}.$(date +%-m%d)"
 
+    if [ "${DRY_RUN}" = true ]; then
+        git_opt="--dry-run"
+    else
+        git_opt=""
+    fi
+
     echo ">>> [beta ${drive_version}] Creating the commit"
     change_version "${drive_version}"
     git commit -m "Release $drive_version"
-    if [ "${DRY_RUN}" = true ]; then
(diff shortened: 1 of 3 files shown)


tools/release.sh
```diff
@@ -65,41 +67,67 @@ create_beta() {
     echo ">>> [beta ${drive_version}] Creating the commit"
     change_version "${drive_version}"
     git commit -m "Release $drive_version"
-    git push origin master
+    if [ "${DRY_RUN}" = true ]; then
+        echo "DRY-RUN git push origin master"
+        git log -n 5 --abbrev-commit --pretty=oneline
```
ataillefer: Could use the `-n, --dry-run` option for `git push`

tools/release.sh
```diff
@@ -65,41 +67,67 @@ create_beta() {
...
     changelog "${drive_version}"
 
     echo ">>> [beta ${drive_version}] Creating the tag"
-    git tag "release-${drive_version}"
-    git push origin --tags
+    if [ "${DRY_RUN}" = true ]; then
+        echo "DRY-RUN git tag \"release-${drive_version}\""
+        echo "DRY-RUN git push origin --tags"
+    else
+        git tag "release-${drive_version}"
```
ataillefer: You can tag even in dry-run as long as you don't push the tag

tools/release.sh
```diff
@@ -65,41 +67,67 @@ create_beta() {
...
+        git push origin --tags
```
ataillefer: If DRY_RUN: `git push origin --tags --dry-run`

tools/release.sh
BoboTiG: Of course.

tools/release.sh
BoboTiG: Done.

PR #206, opened 2017-04-13
Try to use repeat=5 instead of BYPASS

ataillefer:


BoboTiG:
Will be retried later. I close as too many conflicts and will open a new one.

PR #220, opened 2017-04-26
NXDRIVE-610: Optimize logging configuration

BoboTiG:
Tests are OK en GNU/Linux and Windows. Timeout for Mac but changes are trivial.


loopingz:
👍

PR #232, opened 2017-05-05
NXDRIVE-645-686 Fixes

BoboTiG:
NXDRIVE-645 : Add the test case (fixed by NXDRIVE-822)
NXDRIVE-686: Add several tests + fix (thanks to @mkeshava)

Tests were green but we lost the trace with the Jenkins restart ...

### DIFF
--- a/nuxeo-drive-client/tests/test_shared_folders.py
+++ b/nuxeo-drive-client/tests/test_shared_folders.py
@@ -178,3 +178,50 @@ class TestSharedFolders(UnitTestCase):
             # File has been renamed and image deleted
             self.assertEqual(file_info.name, 'File01_renamed.txt')
             self.assertFalse(remote_2.exists('/Folder01/SubFolder01/Image01.png'))
+
+    def test_conflict_resolution_with_renaming(self):
+        """ NXDRIVE-645: shared Folders conflict resolution with renaming. """
+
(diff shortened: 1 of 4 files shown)


nuxeo-drive-client/tests/test_shared_folders.py
```diff
@@ -178,3 +178,50 @@ def _test_local_changes_while_not_running(self, unbind):
...
+        self.assertTrue(remote_2.exists('/Finished/Aerial04.jpg'))
+
+        # Check client 1
+        self.assertTrue(local_1.exists('/Finished'))
+        self.assertTrue(local_1.exists('/Finished/Aerial04.jpg'))
+
+        # Check folder status
+        folder_pair_state = self.engine_1.get_dao().get_state_from_id(
+            folder_pair_state.id)
+        self.assertEqual(folder_pair_state.last_error, 'DEDUP')
```
ataillefer: Could check 'conflicted' state too

BoboTiG: In fact, the state will be `remotely_modified` or `remotely_created` but not `conflicted`. The file is not in conflict, it is in error only.

ataillefer: :+1:

PR #234, opened 2017-05-24
NXDRIVE-849: Handle remote rename when doing full remote scan

BoboTiG:
There are a lot of changes here:
    - NXDRIVE-849: fix
    - NXDRIVE-846: removal of several debug log lines
    - NXDRIVE-626: use of Crowdin (this is mostly a preparation, I sorted locales and translations sentences everywhere and review translator.py)
    - I generalised workers by using `**kwargs` to prevent to have to modify their signature for future changes
    - set `log-level-file` to `DEBUG` as it used to generate too many informations
    - PEP8 and cleanup as always

You can find more informations about 849 in the commit associated message.

Finlly, tests are green and syncing about 3,000 documents while locally and remotely renaming folders work without any error.


nuxeo-drive-client/tests/test_remote_move_and_rename.py
```diff
@@ -268,7 +268,6 @@ def suspend_check(reason):
         self.assertFalse(local.exists('/Test folder/testFile.pdf'))
         self.assertTrue(local.exists('/Test folder/New folder/testFile.pdf'))
 
-    @RandomBug('NXDRIVE-808', target='linux', repeat=2)
     def test_synchronize_remote_rename_while_download_file(self):
```
ataillefer: Is this test enough to cover the scenario that led to NXDRIVE-849?

nuxeo-drive-client/nxdrive/data/ui5/activity.html
```diff
@@ -21,7 +21,7 @@
 </script>
 </head>
 <body style="overflow: hidden;" ng-app="DriveActivity" ng-controller="DriveActivityCtl">
-	<div drive-include="getTemplate('activity-content')">
+	<div drive-include="getTemplate('activity-content.html')">
```
loopingz: The problem isnt more the fact that the file in template does not contain the .html extension ?

nuxeo-drive-client/nxdrive/engine/dao/sqlite.py
```diff
@@ -571,12 +570,7 @@ def release_processor(self, processor_id):
                 con.commit()
         finally:
             self._lock.release()
-        res = c.rowcount > 0
-        if res:
-            log.trace('Released processor %d', processor_id)
-        else:
-            log.trace('No processor to release with id %d', processor_id)
-        return res
+        return c.rowcount > 0
```
loopingz: This will break the log analyser

nuxeo-drive-client/nxdrive/wui/translator.py
```diff
@@ -1,85 +1,78 @@
...
+        if not values:
             return label
+
         result = label
-        for token in re.findall(r'\{\{[^\}]+\}\}', label):
+        for token in re.findall(r'{{[^}]+}}', label):
             attr = token[2:-2].strip()
-            value = ""
-            if attr in values:
+            try:
```
loopingz: Why using an exception instead of just testing ?

nuxeo-drive-client/nxdrive/data/ui5/activity.html
BoboTiG: Yes, you are right. I reverted this change.

nuxeo-drive-client/nxdrive/wui/translator.py
BoboTiG: I prefer the idiomatic EAFP (Easier to Ask for Forgiveness than Permission) in general. Here, it will prevent twice lookups for instance.

PR #239, opened 2017-06-15
NXDRIVE-881: Handle folder remote rename while syncing

BoboTiG:
There was a misuse (or perhaps an oversight) with the `mark_unknown` keyword. I deleted it and added a check for remote and local paths to determine if there was a change. If so, update DB states and add the new path for future scan.

Tested with 1,000, 2,000 and +3,000 files: OK.

I open the PR now for review but will wait for the server fix before merging. We need to test at 100% that change.


nuxeo-drive-client/nxdrive/engine/watcher/remote_watcher.py
```diff
@@ -164,24 +164,33 @@ def _scan_pair(self, remote_path):
...
+            child_pair.remote_can_rename != child_info.can_rename,
+            child_pair.remote_can_update != child_info.can_update,
+            child_pair.remote_can_create_child != child_info.can_create_child,
+            child_pair.remote_digest != child_info.digest,
+        ])
+
+    def _do_scan_remote(self, doc_pair, remote_info, force_recursion=True, moved=False):
+        # NXDRIVE-881: check the folder was not remotely renamed.
+        # Note: local_name = '' for the root folder, so skip it.
+        local_name = os.path.basename(doc_pair.local_path)
```
ataillefer: Use `doc_pair.local_name`?

nuxeo-drive-client/nxdrive/engine/watcher/remote_watcher.py
```diff
@@ -164,24 +164,33 @@ def _scan_pair(self, remote_path):
...
+        if local_name and local_name != remote_info.name:
+            self._dao.force_remote(doc_pair)
```
ataillefer: This will also set `local_state='synchronized'`, could it be an issue?

nuxeo-drive-client/nxdrive/engine/watcher/remote_watcher.py
```diff
@@ -164,24 +164,33 @@ def _scan_pair(self, remote_path):
...
+            self._dao.add_path_to_scan(
+                self._init_scan_remote(doc_pair, remote_info))
```
ataillefer: Why not just `self._dao.add_path_to_scan(remote_info.path)`? I think that `_init_scan_remote` will then always be called in either `_scan_remote_scroll` or `_scan_remote_recursive`, though it might be interesting to check at this point, not sure...

BoboTiG:
Changes applyied. Wait and see!

PR #243, opened 2017-06-22
NXDRIVE-680: Fix unwanted local upload on network connection lost

BoboTiG:
Also:
    - Fixed 3 regressions introduced with NXDRIVE-858.

### DIFF
--- a/nuxeo-drive-client/tests/test_synchronization_suspend.py
+++ b/nuxeo-drive-client/tests/test_synchronization_suspend.py
@@ -1,13 +1,8 @@
-from nxdrive.client import LocalClient
+# coding: utf-8
 from tests.common_unit_test import RandomBug, UnitTestCase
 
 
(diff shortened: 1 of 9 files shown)


nuxeo-drive-client/tests/test_synchronization_suspend.py
```diff
@@ -32,6 +27,46 @@ def test_basic_synchronization_suspend(self):
...
+        # Checks
+        self.assertTrue(remote.exists('/file1.txt'))
+        self.assertTrue(local.exists('/file1.txt'))
+
+        # Simulate offline mode (no more network for instance)
+        engine.get_queue_manager().suspend()
+
+        # Create a bunch of files locally
+        local.make_folder('/', 'files')
+        for num in range(20):
```
ataillefer: Shouldn't we check with more than 50 files, value of the `_windows_queue_threshold` that triggers the full scan?

PR #282, opened 2017-08-30
NXDRIVE-961: Add auto-update process check script

BoboTiG:


ataillefer:
Nice! How do you actually check that the emulated old version updates to the newer version?

BoboTiG:
When Drive is launched, you can see traces and it will restart using the new version. Or not, depending an eventual regression. All simply :)

BoboTiG:
I think we can automate the thing and check `ndrive --version`. We will need a way to quit Drive after the update check is done. But for now, it works fine.

ataillefer:
:+1:

PR #323, opened 2017-10-24
NXDRIVE-998: Fix Drive launch if the pac URL is not reachable

rajadg:


nuxeo-drive-client/nxdrive/manager.py
```diff
@@ -232,13 +232,13 @@ def get_proxies_automatic(self, url):
...
 
-        default = {}
+        default = None
         if self.config != 'Automatic':
             return default
 
         try:
             response = urllib2.urlopen(self.pac_url)
-        except urllib2.HTTPError:
+        except urllib2.URLError:
```
rajadg: The urlopen method throws URLError when pac script is not available.

nuxeo-drive-client/nxdrive/manager.py
```diff
@@ -232,13 +232,13 @@ def get_proxies_automatic(self, url):
                 for given URL.
         """
 
-        default = {}
+        default = None
```
rajadg: The old code was returning {} and this is interpreted as no proxy (direct connection) and so even after PAC url is available drive client did not retrieve PAC url. 
So now the get_proxies_automatic method returns 'None'. When PAC url is not found, then drive client will make future attempts at resolving the PAC url. As soon as PAC url is available drive client will automatically go online.

nuxeo-drive-client/nxdrive/manager.py
```diff
@@ -1063,7 +1063,7 @@ def get_proxies(self, server_url):
         elif proxy_settings.config == 'System':
             proxies = self.retreive_system_proxies(server_url)
         elif proxy_settings.config == 'Automatic':
-            if server_url not in self.proxies:
+            if server_url not in self.proxies or self.proxies[server_url] is None:
```
BoboTiG: You could use `self.proxies.get(server_url) is None` to prevent double checks.

rajadg:
Thanks for the review. I have updated the code as requested.

nuxeo-drive-client/nxdrive/manager.py
rajadg: done.

PR #377, opened 2018-01-03
NXDRIVE-836: Bad behaviors with read-only documents

BoboTiG:
Added a lot of tests to cover all cases.

Also:
  - Framework: Review LocalWatcher class, better use of lock and atomic methods
  - Framework: Deleted the Next Engine source tree
  - Log: Review how TRACE level is added to loggers
  - Tests: Log messages are now less verbose ("thread module message")
  - Updater: Updated minimum server version from 5.6 to 7.10


BoboTiG:
Too broad, will open a new one more focused.

PR #391, opened 2018-01-30
NXDRIVE-887: Integrate SonarCloud code quality check

LKleinNux:


tools/qa.sh
```diff
@@ -0,0 +1,32 @@
...
+
+code_coverage() {
+    echo ">>> [QA] Code coverage"
+    python -m coverage combine
+    python -m coverage xml -i
+}
+
+code_quality() {
+    echo ">>> [QA] Code quality"
+    python -m pylint nuxeo-drive-client/nxdrive > pylint-report.txt || exit 0
```
BoboTiG: It is better to use `... || true`.

tools/qa.sh
```diff
@@ -0,0 +1,32 @@
...
+    virtualenv -p python2 venv
+    . venv/bin/activate
+    python -m pip install coverage pylint
+}
+
+main() {
+    setup
+    code_coverage
+    code_quality
+}
```
BoboTiG: Perhaps a `deactivate` would be a good thing?

tools/jenkins/tests.groovy
```diff
@@ -176,5 +185,56 @@ for (def x in slaves) {
...
+                        dir('sources') {
+                            for (def slave in slaves) {
+                                try {
+                                    unstash "coverage_${slave}"
+                                    sh "mv .coverage .coverage.${slave}"
+                                    echo "Unstashed .coverage.${slave}"
+                                } catch(e) {
+                                    currentBuild.result = 'UNSTABLE'
+                                    echo e
+                                }
```
BoboTiG: In case of error, you can keep `throw e` because the rest of the job will have bad results.

PR #396, opened 2018-01-31
NXDRIVE-1104: Set invalid credentials on 401,403 errors only

BoboTiG:
The systray icon will no longer being red when a HTTP error is not a real credentials one.

### DIFF
--- a/nuxeo-drive-client/tests/test_synchronization.py
+++ b/nuxeo-drive-client/tests/test_synchronization.py
@@ -7,10 +7,17 @@ from nxdrive.client import LocalClient
 from nxdrive.client.remote_filtered_file_system_client import \
     RemoteFilteredFileSystemClient
 from nxdrive.osi import AbstractOSIntegration
-from tests import RemoteTestClient
-from tests.common import OS_STAT_MTIME_RESOLUTION, TEST_WORKSPACE_PATH
-from tests.common_unit_test import DEFAULT_WAIT_SYNC_TIMEOUT, RandomBug, \
+from . import RemoteTestClient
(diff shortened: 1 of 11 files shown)


nuxeo-drive-client/tests/test_synchronization.py
```diff
@@ -7,12 +7,19 @@
 from nxdrive.client.remote_filtered_file_system_client import \
     RemoteFilteredFileSystemClient
 from nxdrive.osi import AbstractOSIntegration
-from tests import RemoteTestClient
-from tests.common import OS_STAT_MTIME_RESOLUTION, TEST_WORKSPACE_PATH
-from tests.common_unit_test import DEFAULT_WAIT_SYNC_TIMEOUT, RandomBug, \
+from . import RemoteTestClient
+from .common import OS_STAT_MTIME_RESOLUTION, TEST_WORKSPACE_PATH
+from .common_unit_test import DEFAULT_WAIT_SYNC_TIMEOUT, RandomBug, \
```
LKleinNux: On the Python Client, I used parentheses around lists of imports to avoid the \ when continuing on another line. Should we keep the same convention? Just a cosmetic question.

nuxeo-drive-client/tests/test_synchronization.py
```diff
@@ -140,6 +147,32 @@ def test_single_quote_escaping(self):
         self.wait_sync(wait_for_async=True)
         self.assertFalse(local.exists("/APPEL D'OFFRES"))
 
+    def test_invalid_credentials(self):
```
LKleinNux: 👌

PR #399, opened 2018-02-07
NXDRIVE-1098: Auto-Lock does not work when there are orphans

BoboTiG:
Also, a lot of cleanup:
- Framework: do not create test folder for case sensitivity in the Drive folder but in a temporary one
- Remove a debug line from InvalidBatchException
- `DirectEdit._url` -> `.url`
- Better logic when using DE if Drive is already running
- Review DirectEdit class
- Add utils.py::`simplify_url()` and refactor test_utils.py to use pytest format
- Removed `LocalClient.make_tree()`

### DIFF
--- a/nuxeo-drive-client/nxdrive/direct_edit.py
+++ b/nuxeo-drive-client/nxdrive/direct_edit.py
@@ -37,17 +37,18 @@ class DirectEdit(Worker):
 
     def __init__(self, manager, folder, url):
         super(DirectEdit, self).__init__()
-        self._test = False
+
         self._manager = manager
-        self.url = url
-        self._thread.started.connect(self.run)
-        self._event_handler = None
-        self._metrics = dict()
-        self._metrics['edit_files'] = 0
-        self._observer = None
         if isinstance(folder, bytes):
             folder = unicode(folder)
         self._folder = folder
+        self.url = url
+
+        self._test = False
+        self._thread.started.connect(self.run)
+        self._event_handler = None
+        self._metrics = {'edit_files': 0}
+        self._observer = None
         self._local_client = LocalClient(self._folder)
         self._upload_queue = Queue()
         self._lock_queue = Queue()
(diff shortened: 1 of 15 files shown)


nuxeo-drive-client/nxdrive/direct_edit.py
```diff
@@ -424,41 +446,55 @@ def _handle_queues(self):
...
                 self._error_queue.push(ref, ref)
-                continue
+
+    def _handle_queues(self):
+        # Lock any document
+        self._handle_lock_queue()
+
+        # Unqueue any errors
+        item = self._error_queue.get()
+        while item:
```
LKleinNux: Why not use a generator?

nuxeo-drive-client/nxdrive/direct_edit.py
```diff
@@ -301,91 +317,96 @@ def edit(self, server_url, doc_id, user=None, download_url=None):
...
                 # Try again in 30s
                 log.exception('Cannot %s document %r', action, ref)
                 self.directEditLockError.emit(action, os.path.basename(ref), uid)
 
-        # Unqueue any errors
-        item = self._error_queue.get()
-        while item:
-            self._upload_queue.put(item.get())
-            item = self._error_queue.get()
+    def _hangle_upload_queue(self):
```
LKleinNux: Typo here

nuxeo-drive-client/nxdrive/direct_edit.py
```diff
@@ -485,70 +520,93 @@ def _stop_watchdog(self):
...
 
+    def _is_regular_file(self, name):
+        # type: (str) -> bool
+        """
+        Check if a given file name is not a temporary one created by
+        a tierce software.  We use it to disable the feature as we
+        use the global open files instead of editor lock file.
+        """
+
+        return False and self._is_lock_file(name)
```
LKleinNux: Wouldn't that always return False?

BoboTiG:
Changes done. Also:
- Review BlackListQueue class (and so tests)

PR #400, opened 2018-02-09
NXDRIVE-1062: Fix macOS warnings

LKleinNux:
- Force unicode in string comparisons
- Add -W error pytest option


nuxeo-drive-client/nxdrive/engine/dao/sqlite.py
```diff
@@ -79,7 +79,7 @@ def cursor(self):
 class StateRow(sqlite3.Row):
 
     def __init__(self, arg1, arg2):
```
BoboTiG: You can safely delete the method.

PR #405, opened 2018-02-12
NXBT-1744 fix discardoldbuild jenkinsfile

alexistimic:
NXBT-1744 fix discardoldbuild jenkinsfile 
The changes made before have not been applied during the build. The Builddiscarder/logrotator classes have to be inline.


jcarsique:
👍

PR #466, opened 2018-04-13
NXDRIVE-1179: Lost focus when FinderSync refreshes icons

LKleinNux:
Getting FinderSync status through socket communication rather than a custom protocol URL

### DIFF
--- a/nxdrive/osi/darwin/darwin.py
+++ b/nxdrive/osi/darwin/darwin.py
@@ -6,9 +6,9 @@ import sys
 import urllib2
 from logging import getLogger
 
-from nxdrive.engine.workers import Worker
 from .. import AbstractOSIntegration
 from ...constants import BUNDLE_IDENTIFIER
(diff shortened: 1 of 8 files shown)


nxdrive/osi/darwin/darwin.py
```diff
@@ -1,13 +1,15 @@
 # coding: utf-8
 import os
+import socket
 import stat
 import sys
 import urllib2
 from logging import getLogger
 
+from nxdrive.engine.workers import Worker
```
BoboTiG: Use the relative import.

nxdrive/wui/application.py
```diff
@@ -654,25 +654,6 @@ def _handle_macos_event(self, url):
         path = info.get('filepath', None)
         manager = self.manager
 
-        # Note: commands are sorted by usage intensity
```
BoboTiG: You can remove unused imports ;)

PR #491, opened 2018-05-17
 NXDRIVE-1078: Create a pytest plugin for random bugs

LKleinNux:


tests/test_remote_file_system_client.py
```diff
@@ -172,7 +173,7 @@ def test_scroll_descendants(self):
         remote = self.remote_file_system_client_1
 
         if 'NuxeoDrive.ScrollDescendants' not in remote.operations:
```
BoboTiG: Can you remove the whole check please? Without this operation, no tests. For now it just eats bandwith to retrieve the operations list.

PR #496, opened 2018-05-23
NXDRIVE-1201: Adapt Drive for new Trash API behavior

LKleinNux:


nxdrive/client/remote_client.py
```diff
@@ -640,7 +645,7 @@ def doc_to_info(self, doc, fetch_parent_uid=True, parent_uid=None):
             self._base_folder_ref, name, doc['uid'], parent_uid,
```
BoboTiG: For preventing future miss and for readability, could you use keywaord argumments?

PR #514, opened 2018-06-29
NXDRIVE-691: Upgrade to Python 3.6.5 (and Qt 5)!

BoboTiG:
A massive one with a lot of changes which includes:

- NXDRIVE-692: Upgrade from PyQt4 to PyQt5
- NXDRIVE-811: Fix test_local_rename_file_while_creating_before_marker on Windows
- NXDRIVE-876: Make NotFound exception inherits from OSError
- NXDRIVE-900: Improve memory consumption
- NXDRIVE-922: Move custom exceptions to exceptions.py
- NXDRIVE-1210: Sanitize exported objects 
- NXDRIVE-1220: Fix select: Invalid argument on macOS
- NXDRIVE-1223: Use a fork of universal-analytics-python that supports Python 3
- NXDRIVE-1225: Be specific about pyobjc requirements 

Also:
- Display installers download links in the changelog

Files in the `wui` folder will be reviewed after the QML job is done.


tests/test_engine_dao.py
```diff
@@ -1,292 +1,239 @@
...
-import unittest
+import time
 
 from nxdrive.engine.dao.sqlite import EngineDAO
-from .common import clean_dir
 
 
-def get_default_db(name='test_engine.db'):
-    return os.path.join(os.path.dirname(__file__), 'resources', name)
+class MockEngineDao(EngineDAO):
```
LKleinNux: 👌

PR #536, opened 2018-07-19
NXDRIVE-1272: Fix settings view

LKleinNux:
Fix the tests impacted by the QML move as well.


nxdrive/gui/view.py
```diff
@@ -315,6 +315,10 @@ def data(self, index: QModelIndex, role: int = TAG_ROLE) -> str:
     def getTag(self, index: int):
         return self.languages[index][0]
 
+    @pyqtSlot(int, result=str)
+    def getName(self, index: int):
```
BoboTiG: Missing return type annotation :)

tests/test_translator.py
```diff
@@ -102,8 +103,8 @@ def test_token(token, result):
 @pytest.mark.parametrize("token", ["ERROR_ON_FILE", "ERROR_ON_FOLDER", "ERROR_ON_BOTH"])
 def test_truncated_paths(token):
```
BoboTiG: This test can be moved to `test_utils.py` as the `Translator` does not handle the short name creation anymore.

nxdrive/utils.py
```diff
@@ -686,6 +686,13 @@ def make_tmp_file(folder: str, content: bytes) -> str:
     return path
 
 
+def short_name(name: Union[bytes, str]) -> str:
```
BoboTiG: Could you add a docstring? something like:
```
Shortening a given `name` for notifications, as the text is limited to 200 characters on Windows:
    https://msdn.microsoft.com/en-us/library/windows/desktop/ee330740(v=vs.85).aspx
This is related to Windows, but we apply the truncation everywhere.
```

PR #545, opened 2018-07-30
NXDRIVE-1231: Add a message box to display fatal errors

BoboTiG:


nxdrive/utils.py
```diff
@@ -424,8 +442,7 @@ def guess_mime_type(filename: str) -> str:
         return mime_type
 
     log.trace(
-        "Could not guess mime type for %r, returing application/octet-stream",
-        filename,
+        "Could not guess mime type for %r, returing application/octet-stream", filename
```
LKleinNux: Typo here, should be "returning"

PR #556, opened 2018-08-02
NXDRIVE-1292: Make the authentication easier to customize

BoboTiG:


nxdrive/gui/dialog.py
```diff
@@ -1,44 +1,40 @@
 # coding: utf-8
```
LKleinNux: This file has nothing to do with a "dialog" anymore, maybe we can change it to `api.py`?

PR #559, opened 2018-08-06
NXDRIVE-1294: Fix send2trash packaging error on macOS

BoboTiG:
And other little fixes.


nxdrive/utils.py
```diff
@@ -239,6 +239,12 @@ def version_compare_client(x: str, y: str) -> int:
 
     from distutils.version import StrictVersion
 
+    # Ignore date based version, they will be traited as normal versions
```
LKleinNux: treated* as

PR #565, opened 2018-08-13
NXDRIVE-1301: Package Nuxeo Drive into one big executable file

BoboTiG:
Ouf! We finally are able to codesign on macOS!


BoboTiG:
This is breaking the custom protocol URL on all Oses. Need to find another solution.

PR #561, opened 2018-08-06
NXDRIVE-1291: Use the user's browser for authentication

LKleinNux:

PR #570, opened 2018-08-14
NXDRIVE-1291: Fix the SIP import for PyQt 5.10 on Windows

BoboTiG:


BoboTiG:
Already done with #561.

PR #579, opened 2018-08-21
NXDRIVE-1312: Improve update information in Systray

LKleinNux:


nxdrive/updater/darwin.py
```diff
@@ -37,14 +37,17 @@ def install(self, filename: str) -> None:
         log.debug("Mounted in %r", mount_dir)
 
         self._backup()
+        self._set_progress(70)
```
BoboTiG: 👍 :)

PR #613, opened 2018-09-06
NXDRIVE-1307: Do not use nxdrive protocol from the FinderSync extension

LKleinNux:
Used sockets instead:
- IBM's BlueSocket in the Swift code,
- QTcpSocket in the Python code.


nxdrive/osi/darwin/darwin.py
```diff
@@ -274,52 +274,47 @@ def _find_item_in_list(lst: List[str], name: str) -> Optional[str]:
...
-        self._sock.close()
-        self._manager.send_sync_status(content)
+        self.port = 50675
+        self.newConnection.connect(self.handle_connection)
+
+    def handle_connection(self) -> None:
+        con: QTcpSocket = self.nextPendingConnection()
+        log.debug("Receiving socket connection for FinderSync event handling")
+        if not con or not con.waitForConnected():
+            log.debug(f"Unable to open FinderSync server socket: {con.errorString()}")
```
BoboTiG: Could be a `log.error()` rather?

PR #627, opened 2018-09-24
 NXDRIVE-1254: Finalize type annotations

LKleinNux:


nxdrive/options.py
```diff
@@ -370,39 +369,6 @@ def server_updater(*args: Any) -> "ServerOptionsUpdater":
     We use this to prevent loading any Drive related stuff and keep
     the possibility to import other classes without anything else needed.
     """
-
```
BoboTiG: The whole `server_updater()` function should be deleted from the file. It can be completely removed and the `ServerOptionsUpdater` class must be used directly when needed.
(+ a line in the `technical_changes.md`)

tests/test_remote_client.py
```diff
@@ -75,7 +75,6 @@ def test_get_fs_info(self):
         fs_item_id = FS_ITEM_ID_PREFIX + "fakeId"
         with pytest.raises(NotFound):
             remote.get_fs_info(fs_item_id)
-        assert not remote.get_fs_info(fs_item_id, raise_if_missing=False)
```
BoboTiG: Why this deletion?

tools/osx/fix_app_qt_folder_names_for_codesign.py
```diff
@@ -89,7 +89,7 @@ def move_contents_to_resources(folder: Path) -> Generator[Path, None, None]:
             yield sibbling
 
 
-def main(args: List[str]) -> int:
+def main(args: List[str]) -> None:
```
BoboTiG: Instead, add a `return 0` at the bottom of the function.

tools/stubs/logging.pyi
```diff
@@ -0,0 +1,22 @@
+class Formatter:
```
BoboTiG: Perhaps add a little file docstring to let developers know this is only to take into account the `.trace()` method we manually add at the runtime.

tests/test_remote_client.py
LKleinNux: We removed the `raise_if_missing` argument from the `get_fs_info()` function. Now, if we call it, it will always raise a `NotFound`, which is already tested in the instruction just above it.

PR #631, opened 2018-09-26
NXDRIVE-1356: Fix removed filters not appearing back when unchecked

LKleinNux:
Having a different class and two remote client instances just for a filter seemed excessive. Now the `Remote` client filters by default, but when populating the filters window, passes an argument to deactivate the filtering.

### DIFF
--- a/nxdrive/client/remote_client.py
+++ b/nxdrive/client/remote_client.py
@@ -322,7 +322,9 @@ class Remote(Nuxeo):
             command="NuxeoDrive.FileSystemItemExists", id=fs_item_id
         )
 
-    def get_fs_children(self, fs_item_id: str, filtered=True) -> List[RemoteFileInfo]:
+    def get_fs_children(
+        self, fs_item_id: str, filtered: bool = True
+    ) -> List[RemoteFileInfo]:
         children = self.operations.execute(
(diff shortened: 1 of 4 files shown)


nxdrive/client/remote_client.py
```diff
@@ -322,11 +322,21 @@ def fs_exists(self, fs_item_id: str) -> bool:
             command="NuxeoDrive.FileSystemItemExists", id=fs_item_id
         )
 
-    def get_fs_children(self, fs_item_id: str) -> List[RemoteFileInfo]:
+    def get_fs_children(self, fs_item_id: str, filtered=True) -> List[RemoteFileInfo]:
```
BoboTiG: Add a line in `technical_changes.md` :)

nxdrive/client/remote_client.py
```diff
@@ -609,19 +621,3 @@ def _get_trash_condition(self) -> str:
         if not self._has_new_trash_service:
             return "AND ecm:currentLifeCycleState != 'deleted'"
         return "AND ecm:isTrashed = 0"
-
-
-class FilteredRemote(Remote):
```
BoboTiG: Add a line in `technical_changes.md`.

PR #633, opened 2018-09-28
NXDRIVE-1364: Account window UI information is invalid

LKleinNux:


nxdrive/client/remote_client.py
```diff
@@ -619,6 +619,15 @@ def set_proxy(self, proxy: Proxy = None) -> None:
             settings = proxy.settings(url=self.client.host)
             self.client.client_kwargs["proxies"] = settings
 
+    def get_server_configuration(self) -> Dict[str, Any]:
```
BoboTiG: Add a line in `technical_changes.md` :)

PR #656, opened 2018-11-02
NXDRIVE-1407: Display CLI arguments in the fatal error screen

BoboTiG:


BoboTiG:
[View issue in JIRA](https://jira.nuxeo.com/browse/NXDRIVE-1407)

PR #666, opened 2018-11-09
NXDRIVE-1259: Be more consistent with file/folder renaming

LKleinNux:


nxdrive/utils.py
```diff
@@ -406,6 +407,13 @@ def safe_filename(
     return re.sub(pattern, replacement, name)
 
 
+def safe_os_filename(name: str):
```
BoboTiG: Could you add a docstring and return type? Also you can explain why the diffrence between Windows et Linux/macOS.

PR #678, opened 2018-11-20
NXDRIVE-1417: Handle null properties in objects instantation

LKleinNux:


nxdrive/objects.py
```diff
@@ -141,86 +146,50 @@ class NuxeoDocumentInfo:
...
-        if lock_created is not None:
+        if lock_created:
             lock_created = parser.parse(lock_created)
 
         # Permissions
         permissions = doc.get("contextParameters", {}).get("permissions", None)
 
         # Trashed
-        is_trashed = doc.get("isTrashed", doc["state"] == "deleted")
+        is_trashed = doc.get("isTrashed", doc.get("state") == "deleted")
```
BoboTiG: Perhaps using `doc.get("isTrashed", doc.get("state", "") == "deleted")` to prevent compring `None == "deleted"`?

PR #679, opened 2018-11-20
 NXDRIVE-1430: Use SHA256 for Note checksum

LKleinNux:
NXDRIVE-1431: Replace Try, Except, Pass with contextlib.suppress()

### DIFF
--- a/nxdrive/client/local_client.py
+++ b/nxdrive/client/local_client.py
@@ -613,12 +613,9 @@ FolderType=Generic
             send2trash(os_path)
         except OSError as exc:
             log.error(f"Cannot trash {os_path!r}")
-            try:
+            with suppress(Exception):
                 # WindowsError(None, None, path, retcode)
                 _, _, _, retcode = exc.args
-            except:
-                pass
-            else:
                 exc.winerror = retcode  # type: ignore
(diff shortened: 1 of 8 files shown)


nxdrive/client/local_client.py
```diff
@@ -613,12 +613,9 @@ def delete(self, ref: str) -> None:
             send2trash(os_path)
         except OSError as exc:
             log.error(f"Cannot trash {os_path!r}")
-            try:
+            with suppress(WindowsError):
```
BoboTiG: `WindowsError` may not be the good exception to skip. We can have an `ValueError`, `TypeError` or even `AttributeError`. Perhaps filtering on `Exception` is safer?

nxdrive/report.py
```diff
@@ -111,10 +112,8 @@ def export_logs(lines: int = constants.MAX_LOG_DISPLAYED) -> Iterator[bytes]:
             try:
                 line = handler.format(record)
             except:
-                try:
+                with suppress(AttributeError):
```
BoboTiG: Hm same here. `UnicodeDecodeError` may be possible too. `Exception` is too broad you think?

PR #693, opened 2018-11-30
NXDRIVE-1438: Introduce update channels

BoboTiG:
And other changes in the same PR as done when working on the subject.


BoboTiG:
Will open several smallers PRs.

PR #718, opened 2018-12-19
NXDRIVE-1466: Add OS and Architecture dimensions to Analytics

BoboTiG:


nxdrive/utils.py
```diff
@@ -104,6 +106,32 @@ def current_milli_time() -> int:
...
+
+    return f"{calcsize('P') * 8}-bit"
+
+
+def get_current_os() -> str:
+    """ Detect the OS version. """
+
+    device = get_device()
+    if device == "macOS":
+        version = platform.mac_ver()[0]
```
LKleinNux: By doing `from sys import platform`, here `platform`'s value is `darwin`.
I think you probably need `import platform` in this function and to put `from sys import platform` directly in `get_device()`.

PR #760, opened 2019-01-28
NXDRIVE-1523: Handle paths with more than 143 chars on encrypted FS

BoboTiG:
On most GNU/Linux FS, the maximum filename size is 255.
It is worst when using encryptFS which sets the limit to 143.

The patch makes `Processor()` to handle this case (`OSError` 36) and adapt `LocalClient.exists()` to return `False` when encountering that error.


efge:
> On most GNU/Linux FS

FYI the filesystems have nothing to do with GNU.

BoboTiG:
@efge you are totally right. I tried to use GNU/Linux everywhere when lot of people only say Linux; but this time I did too much :)

PR #763, opened 2019-01-28
NXDRIVE-1519: Allow DirectEditing documents with the Folderish facet

espa77:
…llow direct edit on folderish documents

### DIFF
--- a/nxdrive/objects.py
+++ b/nxdrive/objects.py
@@ -170,6 +170,7 @@ class NuxeoDocumentInfo:
             uid = doc["uid"]
             path = doc["path"]
             props = doc["properties"]
+            file_content_data = doc["properties"].get("file:content").get("data")
             name = unicodedata.normalize("NFC", props["dc:title"])
             folderish = "Folderish" in doc["facets"]
             modified = doc["lastModified"]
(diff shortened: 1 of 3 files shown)


nxdrive/objects.py
```diff
@@ -170,6 +170,7 @@ def from_dict(doc: Dict[str, Any], parent_uid: str = None) -> "NuxeoDocumentInfo
             uid = doc["uid"]
             path = doc["path"]
             props = doc["properties"]
+            file_content_data = doc["properties"].get("file:content").get("data")
```
BoboTiG: We can end on such error:
```python
AttributeError: 'NoneType' object has no attribute 'get'
```
To prevent this, you could use something like (also renamed the var for clarity and reusing the previsous `props` to save one dict lookup):
```python
has_data = props.get("file:content", {}).get("data")
```

nxdrive/objects.py
```diff
@@ -178,7 +179,8 @@ def from_dict(doc: Dict[str, Any], parent_uid: str = None) -> "NuxeoDocumentInfo
 
         last_update = parser.parse(modified)
 
-        blobs = {} if folderish else NuxeoDocumentInfo._parse_blobs(props)
+        # check for file:content/data instead of folderish facet NXDRIVE-1519
```
BoboTiG: I would like to keep comments well formatted:
> Check for file:content/data first instead of only the folderish facet (NXDRIVE-1519)

nxdrive/objects.py
```diff
@@ -178,7 +179,8 @@ def from_dict(doc: Dict[str, Any], parent_uid: str = None) -> "NuxeoDocumentInfo
...
+        blobs = NuxeoDocumentInfo._parse_blobs(props) if file_content_data else {}
```
BoboTiG: Rename to `has_data` here.

PR #791, opened 2019-02-18
NXDRIVE-1553: Check the uuid of a document before trashing it

LKleinNux:


nxdrive/engine/processor.py
```diff
@@ -1229,6 +1229,8 @@ def _create_remotely(
             self._lock_readonly(local_parent_path)
 
     def _synchronize_remotely_deleted(self, doc_pair: DocPair) -> None:
+        if self.local.get_remote_id(doc_pair.local_path) != doc_pair.remote_ref:
+            return
```
BoboTiG: Perhaps could we log a warning here?

PR #807, opened 2019-02-26
NXDRIVE-1468: Factorize FS decoration handling cross-platform

LKleinNux:


nxdrive/osi/extension.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _format_response(self, response: str) -> bytes:
+        """ Called on the string to send through the socket. """
+        return force_encode(response)
+
+    def _handle_content(self, content: str) -> Optional[str]:
+        """ Called on the parsed payload, runs the handler associated with the command. """
+        try:
+            data = json.loads(content)
+        except:
```
BoboTiG: Prefer `except Exception`. We will have some clean-up about those "too global" catches :)

nxdrive/osi/extension.py
```diff
@@ -0,0 +1,148 @@
...
+        if not handler:
+            log.debug(f"No handler for the listener command {cmd}")
+            return None
+
+        response = handler(value)
+        return json.dumps(response)
+
+    def get_engine(self, path: Path) -> Optional[Engine]:
+        for engine in self.manager._engines.values():
+            if engine.local_folder not in path.parents:
```
BoboTiG:
```suggestion
            if engine.local_folder in path.parents:
```

nxdrive/osi/extension.py
```diff
@@ -0,0 +1,148 @@
...
+                continue
```
BoboTiG:
```suggestion
                return engine
```

nxdrive/osi/extension.py
```diff
@@ -0,0 +1,148 @@
...
+            return engine
```
BoboTiG:
```suggestion
```

PR #827, opened 2019-03-12
NXDRIVE-1556: Display a warning when activating DEBUG logs about sensitive data

LKleinNux:


nxdrive/commandline.py
```diff
@@ -367,7 +367,7 @@ def parse_cli(self, argv: List[str]) -> Namespace:
         has_command = False
 
         # Pre-configure the logging to catch early errors
-        configure(console_level="DEBUG", command_name="early")
+        configure(console_level="INFO", command_name="early")
```
BoboTiG: This one should be kept at `DEBUG`, this is the time we initialize the logging and to being able to catch option updates.

PR #828, opened 2019-03-13
NXDRIVE-1501: Prompt for confirmation when a user deletes a document locally

LKleinNux:


nxdrive/engine/dao/sqlite.py
```diff
@@ -1292,6 +1292,19 @@ def remove_state(self, doc_pair: DocPair, remote_recursion: bool = False) -> Non
                     condition = self._get_recursive_condition(doc_pair)
                 c.execute("DELETE FROM States " + condition)
 
+    def remove_state_children(
+        self, doc_pair: DocPair, remote_recursion: bool = False
+    ) -> None:
+        with self._lock:
+            con = self._get_write_connection()
+            c = con.cursor()
+            if doc_pair.folderish:
```
BoboTiG: Could we move this condition outside the method and call it only on folderish doc pairs?

PR #842, opened 2019-03-22
NXDRIVE-1552: Pre-release 4.1.0 fixes

BoboTiG:
* NXDRIVE-1596: Set the `channel` option to release by default. Also fix the `channel` option persistence in the database.
* NXDRIVE-1597: Update options of type sequence instead of overwriting them.
* NXDRIVE-1552: Fix deploy script on macOS.
* NXDRIVE-1552: Fix the `check_update_process.py` script.

### DIFF
--- a/nxdrive/options.py
+++ b/nxdrive/options.py
@@ -153,7 +153,7 @@ class MetaOptions(type):
     options: Dict[str, Tuple[Any, str]] = {
         "browser_startup_page": ("drive_browser_login.jsp", "default"),
         "ca_bundle": (None, "default"),
-        "channel": (None, "default"),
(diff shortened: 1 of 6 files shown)


nxdrive/options.py
```diff
@@ -263,6 +261,7 @@ def set(
         `file` and `section` are used for a better exception message.
 
         Any `list` will be converted to a sorted `tuple`.
+        And any `tuple` will be appened to the current sequence instead of erasing old values.
```
BoboTiG:
```suggestion
        And any `tuple` will be appended to the current sequence instead of erasing old values.
```

PR #855, opened 2019-03-28
NXDRIVE-986: Part 1

BoboTiG:
Fixes and improvements before the actual fix.


BoboTiG:
Closing now, it will be part of the bigger PR to come.

PR #861, opened 2019-04-01
NXDRIVE-1611: Add chunk upload options

LKleinNux:
NXDRIVE-1613: Handle None chunk_size during upload

### DIFF
--- a/nxdrive/options.py
+++ b/nxdrive/options.py
@@ -151,7 +151,9 @@ class MetaOptions(type):
         "browser_startup_page": ("drive_browser_login.jsp", "default"),
         "ca_bundle": (None, "default"),
         "channel": ("release", "default"),
-        "chunk_upload": (False, "default"),
+        "chunk_limit": (20, "default"),
+        "chunk_size": (20, "default"),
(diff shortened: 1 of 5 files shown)


nxdrive/options.py
```diff
@@ -430,3 +433,24 @@ class Options(metaclass=MetaOptions):
...
         raise RuntimeError("Instantiation is not allowed.")
+
+
+#
+# Validators
+#
+
+
+def validate_chunk_limit(value: int) -> int:
+    if 0 < value:
```
BoboTiG:
```suggestion
    if value > 0:
```
More readable.

nxdrive/options.py
```diff
@@ -430,3 +433,24 @@ class Options(metaclass=MetaOptions):
...
+        return value
+    raise ValueError("Chunk limit must be above 0")
+
+
+def validate_chunk_size(value: int) -> int:
+    if 0 < value < 20:
```
BoboTiG:
```suggestion
    if 0 < value <= 20:
```

PR #867, opened 2019-04-03
NXDRIVE-1593: Use java 11 for tests

BoboTiG:
* Removed 8.10 jenkinsfile and pom.xml (not used).


tools/jenkins/tests.groovy
```diff
@@ -184,7 +184,7 @@ for (def x in slaves) {
                             //    return
                             //}
 
-                            def jdk = tool name: 'java-8-openjdk'
+                            def jdk = tool name: 'java-11-openjdk'
                             if (osi == 'macOS') {
                                 jdk = tool name: 'java-8-oracle'
```
kevinleturc: Should be changed there too right?

BoboTiG: Yes, updated :)

PR #896, opened 2019-04-16
NXDRIVE-1631: [Windows] Handle files with the same name but different unicode norms

LKleinNux:


nxdrive/engine/watcher/local_watcher.py
```diff
@@ -1176,7 +1177,27 @@ def handle_watchdog_event(self, evt: FileSystemEvent) -> None:
             return
         except ThreadInterrupt:
             raise
-        except:
+        except OSError as exc:
+            if exc.errno == 17:
```
BoboTiG: Does this cover `WindowsError: [Error 183] Cannot create a file when that file already exists`?
Could you also use a named error like `errno.EEXIST`?

LKleinNux: Yes, `OSError` with `errno == 17` is equivalent to `WindowsError: [Error 183]` 👍

PR #900, opened 2019-04-19
NXDRIVE-1623: [Windows] Add CLI sub-commands tests: clean-folder, con…

BoboTiG:
…sole, bind-server and unbind-server

Those sub-commands are not easily testable, so they will not be covered now:

* bind-root (need dynamic UID)
* unbind-root (need dynamic UID)
* access-online (need browser manipulations)
* copy-share-link (need dynamic UID)
* edit-metadata (need browser manipulations)

Changes:

* CLI: renamed `clean_folder` to `clean-folder`.
* Lower several `logging.error()` in `commandline.py` to `logging.warning()`.
* Simplfy `set_path_readonly()` and `unset_path_readonly()` implementations.
* Refactor some test fixtures.

Also including:

* NXDRIVE-1651: Expand environment variables in paths given to CLI arguments. This is cool to be able to pass `--local-folder="%temp%\testing"` for instance (when Windows cannot expand itself the variable).

### DIFF
--- a/nxdrive/commandline.py
+++ b/nxdrive/commandline.py
@@ -9,7 +9,6 @@ from argparse import ArgumentParser, Namespace
 from configparser import DEFAULTSECT, ConfigParser
 from datetime import datetime
 from logging import getLogger
-from pathlib import Path
 from typing import List, TYPE_CHECKING, Union
 
 from . import __version__
@@ -22,6 +21,7 @@ from .utils import (
     get_default_nuxeo_drive_folder,
     get_value,
     normalized_path,
+    normalize_and_expand_path,
 )
 
 try:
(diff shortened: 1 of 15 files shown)


nxdrive/commandline.py
```diff
@@ -453,6 +454,10 @@ def _configure_logger(self, command: str, options: Namespace) -> None:
     def uninstall(self, options: Namespace = None) -> None:
         AbstractOSIntegration.get(None).uninstall()
 
+    def normalize_path(self, path: str) -> Path:
```
LKleinNux: This method could be static, or even a top-level function?

BoboTiG: Yes, and perhaps should we use it on `Options` directly. Like that we could expand variables from others sourcs (config.ini, server config).

PR #914, opened 2019-04-29
NXDRIVE-1647: Check HTTPS when starting an engine

LKleinNux:


nxdrive/engine/engine.py
```diff
@@ -866,6 +867,25 @@ def _check_fs(self, path: Path) -> None:
...
+        if self.server_url.startswith("https"):
+            return
+
+        try:
+            url = self.server_url.replace("http://", "https://")
+            proxies = self.manager.proxy.settings(url=url)
+            import requests
+
+            requests.get(url, proxies=proxies)
+            self.server_url = url
```
BoboTiG: Perhaps should we check for `status_code == 200` only?

LKleinNux: The server could return a `302` because it uses a custom path (instead of `<host>/nuxeo`), a `401` or `403` depending on auth status or request origin, or any number of custom codes for different reasons. Maybe only filter what's equal or above `500` ?

BoboTiG: Hm we want to ensure we are using HTTPS if the server supports it. As we only do a `GET`, perhaps using `r.ok` would be enough? We can't get a `40x` there, else no one would be able to login.

BoboTiG: Or `r.raise_for_status()`?

LKleinNux: We can and probably will get a `40x` there, we do not provide a token at this moment so the server will surely send a `401`.

LKleinNux: If we get back any kind of response, whatever the code is, the server supports HTTPS, otherwise we would get a `ConnectionError` or an `SSLError`.

BoboTiG: :+1:

PR #924, opened 2019-05-10
NXDRIVE-1664: Add OS and Python versions to Sentry events

BoboTiG:


nxdrive/__main__.py
```diff
@@ -234,8 +234,20 @@ def main() -> int:
         setup_sentry()
 
         from nxdrive.commandline import CliHandler
-
-        ret = CliHandler().handle(sys.argv[1:])
+        from nxdrive.utils import get_current_os
+        from sentry_sdk import configure_scope
+
+        with configure_scope() as scope:
```
LKleinNux: Why not put what's in this block directly in the `setup_sentry()` function?

BoboTiG: Because the `context` will be cleared when the context manager is left.

PR #925, opened 2019-05-13
NXDRIVE-1148: Better handle no space left on the device

BoboTiG:
Introduced `NO_SPACE_ERRORS` to handle different kind of disk space related errors.

Different scenarii:

* On database backup: log a warning
* On database restore: raise a fatal error
* On upgrade: alert the user
* On local file opening: raise a fatal error
* Anytime while syncing: alert the user and pause the sync

### DIFF
--- a/nxdrive/constants.py
+++ b/nxdrive/constants.py
@@ -1,4 +1,5 @@
 # coding: utf-8
+import errno
 from enum import Enum
 from pathlib import Path
 from sys import platform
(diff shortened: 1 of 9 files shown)


nxdrive/constants.py
```diff
@@ -30,6 +31,19 @@
 # The registry key from the HKCU hive where to look for local configuration on Windows
 CONFIG_REGISTRY_KEY = "Software\\Nuxeo\\Drive"
 
+# OSError indicating a lack of disk space
+# >>> import errno, os
```
BoboTiG: Not sure if we need to keep this.

nxdrive/engine/processor.py
```diff
@@ -408,10 +408,10 @@ def _execute(self) -> None:
     def _handle_pair_handler_exception(
         self, doc_pair: DocPair, handler_name: str, e: Exception
     ) -> None:
-        if isinstance(e, OSError) and e.errno == errno.ENOSPC:
+        if isinstance(e, OSError) and e.errno in NO_SPACE_ERRORS:
             self.engine.suspend()
-            log.warning("No space left on device!")
-            self.increase_error(doc_pair, "NO_SPACE_LEFT_ON_DEVICE")
+            log.warning("No space left on device!", exc_info=True)
+            self.engine.get_queue_manager().push(doc_pair)
```
BoboTiG: Maybe it is better to keep the old behavior (pushing the doc in error)?

LKleinNux: I think the new behavior is good. It's not the document's fault that there is no space left, so it doesn't need to be in error.

nxdrive/constants.py
LKleinNux: As you wish

PR #932, opened 2019-05-15
NXDRIVE-1649: Review connection errors handling and logging

LKleinNux:


nxdrive/direct_edit.py
```diff
@@ -408,6 +414,9 @@ def _prepare_edit(
             if tmp_file is None:
                 log.warning("Download failed")
                 return None
+        except CONNECTION_ERROR:
+            log.warning("Unable to perform direct edit", exc_info=True)
```
BoboTiG:
```suggestion
            log.warning("Unable to perform DirectEdit", exc_info=True)
```

nxdrive/engine/engine.py
```diff
@@ -9,6 +9,7 @@
 from typing import Any, Callable, Dict, List, Optional, Type, TYPE_CHECKING
 from urllib.parse import urlsplit
 
+import requests
```
BoboTiG: Maybe it is worth replacing this import in `_check_https()` only when we need it?

nxdrive/engine/processor.py
```diff
@@ -296,6 +301,9 @@ def _execute(self) -> None:
                         self._postpone_pair(doc_pair, "Conflict")
                     elif exc.status == 500:
                         self.increase_error(doc_pair, "SERVER_ERROR", exception=exc)
+                    elif exc.status in {502, 503}:
+                        log.warning(f"Server is unavailable", exc_info=True)
```
BoboTiG:
```suggestion
                        log.warning("Server is unavailable", exc_info=True)
```

PR #986, opened 2019-06-14
NXDRIVE-980: Renaming files server-side to solve duplicate issues results in wrong local files count

LKleinNux:


nxdrive/engine/watcher/remote_watcher.py
```diff
@@ -870,6 +870,29 @@ def _update_remote_states(self) -> None:
                         remote_parent_path = os.path.dirname(new_info.path)
                         # TODO Add modify local_path and local_parent_path
                         # if needed
+                        if doc_pair.pair_state == "remotely_created" and (
+                            not doc_pair.local_path.exists()
+                            or self.engine.local.get_remote_id(doc_pair.local_path)
+                            != doc_pair.remote_ref
+                        ):
+                            # We are trying to synchronize a duplicate that has been renamed remotely
+                            from nxdrive.client.local_client import FileInfo
```
BoboTiG: Could you use the relative import and move it to the top of the file?

PR #1015, opened 2019-07-09
NXDRIVE-1219: Use in-memory files instead of physical temporary ones

BoboTiG:


BoboTiG:
Superseeded by #1021.

Issue #541, opened 2018-07-23
sync is slow


tamasberesoebb:
I'm using the 3.1.1 client, and it's copying files from nuxeo to my hard drive at 4-6 Files/second.
I'm using the mac client on a 12 core 64gb imac pro (high sierra 10.13.6) , with a 70 Megabyte/second connection to our server.
I have 110000 ~300kb files on our server (< 100Gb total file size). The synch would take many days to finish.

Back when I only had thousands of only big (>500Mb) files (30Gb total size) it took less than 30 minutes to synch everything between the nuxeo server and the nuxeo synch client running on my mac.

How could I improve the sync speed for a lot of small files?

BoboTiG:
Do you mind retry with the latest beta 4.0.0? The download link is at the bottom of https://github.com/nuxeo/nuxeo-drive/releases/tag/release-4.0.0.

tamasberesoebb:
Sorry, I can't.

"Your server version is not compatible with Drive 4.0.0. We advise you to downgrade your Drive client to version 3.1.2 or to contact your administrator to upgrade your Nuxeo instance."

PR #1127, opened 2019-10-08
NXDRIVE-1839: The GUI should be responsive when syncing a lot of files

BoboTiG:
Added timers to control when the files list in the systray can be updated. Same for syncing docs and synced docs counts.

Some refactoring in `view.py` on how rows insertion and deletion are done. The new code is cleaner and less resources greedy.

There is a small downside though: the remaining docs to sync (at the last iteration) is not updated
until the sync is actually over. But this is another issue not related to the subject and should be tackled another time.

Also:

- NXDRIVE-1875: Use more processors by default (5 -> 10)
- NXDRIVE-1876: Fix threads not totally released


BoboTiG:
Should fix #541 as it is now faster to sync more files. But there is no magic, network can still be a bottleneck.

PR #1156, opened 2019-10-28
NXDRIVE-1897: The clean-folder CLI argument should ignore errors

BoboTiG:
Added a test file mostly to only check the methods and it will also be usefull to validate NXDRIVE-988.


BoboTiG:
[View issue in JIRA](https://jira.nuxeo.com/browse/NXDRIVE-1897)
[View issue in JIRA](https://jira.nuxeo.com/browse/NXDRIVE-988)

PR #1271, opened 2019-12-31
NXDRIVE-1991: Delete old JUnit and pytest caches at tests startup

BoboTiG:


BoboTiG:
Bad patch, the cause is elsewhere.

PR #1349, opened 2020-02-10
NXDRIVE-1831: Set the root local folder icon on Linux

rom1win:
- Updated the set_folder_icon on linux. It now use the gio set command to change the folder custom-icon metadata.

- For the moment, the folder_windows icon is used as there is no icon for linux in the project.

Signed-off-by: Romain Grasland <rgrasland@nuxeo.com>

### DIFF
--- a/nxdrive/client/local/linux.py
+++ b/nxdrive/client/local/linux.py
@@ -27,19 +27,20 @@ class LocalClient(LocalClientMixin):
 
     def has_folder_icon(self, ref: Path) -> bool:
         """Check if the folder icon is set."""
-        matcher = re.compile(r"metadata::emblems: \[nuxeo.*\]")
-        folder_path = self.abspath(ref).as_posix()
+        nuxeo_emblem_path = Path.home() / ".local/share/icons/emblem-nuxeo.svg"
+        if not nuxeo_emblem_path.is_file():
+            return False
+
         try:
             output = subprocess.check_output(
-                ["gio", "info", "-a", "metadata", folder_path], encoding="utf-8"
+                ["gio", "info", "-a", "metadata", self.abspath(ref)], encoding="utf-8"
             )
         except subprocess.CalledProcessError:
             log.debug(f"Could not check the metadata of {ref!r}")
             return False
-        result = matcher.findall(output)
-        if result:
-            return True
-        return False
+
+        matcher = re.compile(r"metadata::emblems: \[nuxeo.*\]")
+        return bool(matcher.findall(output))
 
     @staticmethod
     def get_path_remote_id(path: Path, name: str = "ndrive") -> str:
@@ -69,16 +70,14 @@ class LocalClient(LocalClientMixin):
         """Use commandline to customize the folder icon."""
         log.debug(f"Setting the folder icon of {ref!r} using {icon!r}")
 
-        shared_icons_dir = Path("~/.local/share/icons/")
-        folder_path = self.abspath(ref).as_posix()
-
         # Create the shared icons folder if it doesn't already exist
+        shared_icons_dir = Path.home() / ".local/share/icons"
         shared_icons_dir.mkdir(parents=True, exist_ok=True)
 
         # Emblems icons must be saved in ~/.local/share/icons/ to be accessible
         try:
             emblem_path = shared_icons_dir / "emblem-nuxeo.svg"
-            if emblem_path.is_file():
+            if not emblem_path.is_file():
                 shutil.copy(str(icon), str(emblem_path))
         except shutil.Error:
             log.debug(f"Could not copy {icon!r} to {shared_icons_dir!r}")
@@ -91,14 +90,14 @@ class LocalClient(LocalClientMixin):
                     "set",
                     "-t",
                     "stringv",
-                    folder_path,
+                    self.abspath(ref),
                     "metadata::emblems",
                     "nuxeo",
                 ],
                 encoding="utf-8",
             )
         except subprocess.CalledProcessError:
-            log.debug(f"Could not set the folder icon of {ref!r} using nuxeo emblem")
+            log.warning(f"Could not set the folder emblem on {ref!r}")
 
     @staticmethod
     def set_path_remote_id(
--- a/nxdrive/client/local/linux.py
+++ b/nxdrive/client/local/linux.py
@@ -4,6 +4,7 @@
 import errno
 import os
 import re
+import shutil
 import subprocess
 import unicodedata
 from logging import getLogger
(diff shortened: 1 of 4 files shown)


nxdrive/client/local/linux.py
```diff
@@ -55,6 +56,9 @@ def set_folder_icon(self, ref: Path, icon: Path) -> None:
         """Create a special file to customize the folder icon."""
         log.debug(f"Setting the folder icon of {ref!r} using {icon!r}")
         # To be implementation with https://jira.nuxeo.com/browse/NXDRIVE-1831
```
BoboTiG: You can remove that line.

nxdrive/client/local/linux.py
```diff
@@ -55,6 +56,9 @@ def set_folder_icon(self, ref: Path, icon: Path) -> None:
...
+        subprocess.check_output(
```
BoboTiG: Could you add this code inside a try/except block to prevent crashes.In case of error, log a warning.

Do you know if that will work when there is a space or `'` in those paths?

BoboTiG: Also, have a look at the [macOS version](https://github.com/nuxeo/nuxeo-drive/blob/9102456ca8ce541439eb1948d0846e2967564c41/nxdrive/client/local/darwin.py#L66) and [Windows version](https://github.com/nuxeo/nuxeo-drive/blob/9102456ca8ce541439eb1948d0846e2967564c41/nxdrive/client/local/windows.py#L122) to keep the same flow (logging mostly).

nxdrive/client/local/linux.py
```diff
@@ -24,8 +26,19 @@ class LocalClient(LocalClientMixin):
...
+        matcher = re.compile(r"metadata::emblems: \[nuxeo.*\]")
+        folder_path = self.abspath(ref).as_posix()
+        try:
+            output = subprocess.check_output(
+                ["gio", "info", "-a", "metadata", folder_path], encoding="utf-8"
+            )
+        except subprocess.CalledProcessError:
+            log.debug(f"Could not check the metadata of {ref!r}")
+            return False
+        result = matcher.findall(output)
```
BoboTiG: You could move `matcher = ...` at this line, as it will be useless if the `subprocess` call fails.
Also, you can refactor those lines:
```suggestion
        return bool(matcher.findall(output))
```

(and add a empty line before the whole block to make clear this is not the same "action" being done)

nxdrive/client/local/linux.py
```diff
@@ -24,8 +26,19 @@ class LocalClient(LocalClientMixin):
 
     def has_folder_icon(self, ref: Path) -> bool:
         """Check if the folder icon is set."""
-        # To be implementation with https://jira.nuxeo.com/browse/NXDRIVE-1831
-        return True
+        matcher = re.compile(r"metadata::emblems: \[nuxeo.*\]")
```
BoboTiG: The whole method could be simplified by check cheking the presence of `~/.local/share/icons/emblem-nuxeo.svg`, WDYT?

nxdrive/client/local/linux.py
```diff
@@ -52,10 +65,24 @@ def remove_remote_id_impl(path: Path, name: str = "ndrive") -> None:
...
+                    "-t",
+                    "stringv",
+                    folder_path,
+                    "metadata::emblems",
+                    "nuxeo",
+                ],
+                encoding="utf-8",
+            )
+        except subprocess.CalledProcessError:
+            log.debug(f"Could not set the folder icon of {ref!r} using nuxeo emblem")
```
BoboTiG:
```suggestion
            log.warning(f"Could not set the folder emblem on {ref!r}")
```

nxdrive/engine/engine.py
```diff
@@ -1079,13 +1078,11 @@ def _set_root_icon(self) -> None:
...
 
         if LINUX:
-            # To be implementation with https://jira.nuxeo.com/browse/NXDRIVE-1831
-            return
+            icon = find_icon("disabled.svg")
         elif MAC:
             icon = find_icon("folder_mac.dat")
         else:
             icon = find_icon("folder_windows.ico")
-
```
BoboTiG: Spaces are vital for readability ;)

nxdrive/client/local/linux.py
```diff
@@ -52,10 +65,24 @@ def remove_remote_id_impl(path: Path, name: str = "ndrive") -> None:
                 raise exc
 
     def set_folder_icon(self, ref: Path, icon: Path) -> None:
-        """Create a special file to customize the folder icon."""
+        """Use commandline to customize the folder icon."""
         log.debug(f"Setting the folder icon of {ref!r} using {icon!r}")
-        # To be implementation with https://jira.nuxeo.com/browse/NXDRIVE-1831
-        return
+        folder_path = self.abspath(ref).as_posix()
```
BoboTiG: On GNU/Linux, `.as_posix()` is not needed. You could then move that variable directly into the `subprocess` call.

nxdrive/client/local/linux.py
```diff
@@ -24,8 +26,19 @@ class LocalClient(LocalClientMixin):
 
     def has_folder_icon(self, ref: Path) -> bool:
         """Check if the folder icon is set."""
-        # To be implementation with https://jira.nuxeo.com/browse/NXDRIVE-1831
-        return True
+        matcher = re.compile(r"metadata::emblems: \[nuxeo.*\]")
+        folder_path = self.abspath(ref).as_posix()
```
BoboTiG: On GNU/Linux, `.as_posix()` is not needed. You could then move that variable directly into the `subprocess` call.

nxdrive/client/local/linux.py
```diff
@@ -67,7 +68,22 @@ def remove_remote_id_impl(path: Path, name: str = "ndrive") -> None:
     def set_folder_icon(self, ref: Path, icon: Path) -> None:
         """Use commandline to customize the folder icon."""
         log.debug(f"Setting the folder icon of {ref!r} using {icon!r}")
+
+        shared_icons_dir = Path("~/.local/share/icons/")
```
BoboTiG: You have 2 options:
```python
shared_icons_dir = Path.home() / ".local/share/icons"
```

or
```python
shared_icons_dir = Path("~/.local/share/icons/").expanduser()
```

nxdrive/client/local/linux.py
```diff
@@ -67,7 +68,22 @@ def remove_remote_id_impl(path: Path, name: str = "ndrive") -> None:
...
         folder_path = self.abspath(ref).as_posix()
+
+        # Create the shared icons folder if it doesn't already exist
+        shared_icons_dir.mkdir(parents=True, exist_ok=True)
+
+        # Emblems icons must be saved in ~/.local/share/icons/ to be accessible
+        try:
+            emblem_path = shared_icons_dir / "emblem-nuxeo.svg"
+            if emblem_path.is_file():
```
BoboTiG: This should not be necessary as the check should be done in `has_folder_icon()`.
Here you should just to the copy without checks.

nxdrive/client/local/linux.py
rom1win: If ~/.local/share/icons/emblem-nuxeo.svg exists but not the attribute on the Drive folder then the emblem wont't show. In my opinion we shoul add the file verification before the regexp and return False if the file doesn't exist.

PR #1361, opened 2020-02-17
NXDRIVE-2047: Unlock updates on the centralized channel when auto-upd…

BoboTiG:
…ate is disabled

Auto-updates are now allowed/forced when those parameters are met:

  - the application is frozen;
  - the `update_check_delay` option is set to `0`;
  - the `channel` option is set to `"centralized"`;
  - the `client_version` option is set.

This is required for IT teams wanting to fully control which version of the application to deploy everywhere.


BoboTiG:
This is the first patch out of 3. I would like to merge it before others to be able to work on a clean code base.

PR #1430, opened 2020-03-04
NXDRIVE-2010: Restart the download on HTTP 416 error (range not satis…

rom1win:
…fiable)

When a download is paused, it may happen that the file size on the server
is modified. In this particular case, an HTTP 416 error will be raised.
We want the local temporary file to be deleted and the download to be
restarted.
The processor.py file have been updated to catch and process the error.
A new test have been added in the test_synchronization.py file.
Also the changelog have been updated.

### DIFF
--- a/tests/old_functional/test_synchronization.py
+++ b/tests/old_functional/test_synchronization.py
@@ -819,7 +819,7 @@ class TestSynchronization(OneUserTest):
 
         # Checks
         assert not engine.dao.get_errors()
-        assert self.local_1.exists("/test.bin")
+        assert local.exists("/test.bin")
 
     def test_local_modify_offline(self):
         local = self.local_1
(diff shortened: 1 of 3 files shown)


tests/old_functional/test_synchronization.py
```diff
@@ -791,6 +789,38 @@ def _raise_for_second_file_only(*args, **kwargs):
...
+            self.wait_sync(fail_if_timeout=False)
+            # Checks
+            assert engine.dao.queue_manager.get_errors_count() == 1
+
+        # Starting here, default behavior is restored
+        self.wait_sync()
+
+        # Checks
+        assert not engine.dao.get_errors()
+        assert self.local_1.exists("/test.bin")
```
BoboTiG:
```suggestion
        assert local.exists("/test.bin")
```

PR #1473, opened 2020-03-17
NXDRIVE-1866: The conflict resolver should not fail when the Engine i…

BoboTiG:
…s not started


BoboTiG:
Will reopen a more up-to-date version of the patch.

PR #1512, opened 2020-03-27
NXDRIVE-2074: Always raise the Conflicts/Errors window

rom1win:
When clicking on "Open window" link in the Account settings, the window
may not appear if is already shown and hidden by another application.
A fix has been added in api.py file to always arise the conflict
windows when clicking on the link.

### DIFF
--- a/nxdrive/gui/api.py
+++ b/nxdrive/gui/api.py
@@ -338,7 +338,6 @@ class QMLDriveApi(QObject):
     @pyqtSlot(str)
     def show_conflicts_resolution(self, uid: str) -> None:
         self.application.hide_systray()
-
         engine = self._manager.engines.get(uid)
         if engine:
             self.application.show_conflicts_resolution(engine)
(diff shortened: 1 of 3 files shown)


nxdrive/gui/application.py
```diff
@@ -667,6 +667,7 @@ def show_conflicts_resolution(self, engine: Engine) -> None:
         self.refresh_conflicts(engine.uid)
         self._window_root(self.conflicts_window).setEngine.emit(engine.uid)
         self.conflicts_window.show()
+        self.conflicts_window.raise_()
         self.conflicts_window.requestActivate()
```
BoboTiG: Refactor for less duplicated code:
```suggestion
        self._show_window(self.conflicts_window)
```

nxdrive/gui/api.py
```diff
@@ -338,6 +338,7 @@ def open_help(self) -> None:
     @pyqtSlot(str)
     def show_conflicts_resolution(self, uid: str) -> None:
         self.application.hide_systray()
+
```
BoboTiG: You can revert that change then.

PR #1536, opened 2020-04-02
NXDRIVE-2079: Special case handling when testing the auto-updater

BoboTiG:
Our packaging job tests the auto-updater before validating an alpha release. For that, it does not use an account and bypass that check. But with d8bf965 the auto-updater is started only
after the server's config is fetched, which cannot happen without account.

So we add a special case check to make the alpha release job happy again.

### DIFF
--- a/nxdrive/manager.py
+++ b/nxdrive/manager.py
@@ -1,4 +1,5 @@
 # coding: utf-8
+import os
 import platform
 import shutil
 import sqlite3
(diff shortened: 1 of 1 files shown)


BoboTiG:
Real test to be done tomorrow morning as our agents are all running a long task.

BoboTiG:
Tests passed (https://qa.nuxeo.org/jenkins/view/Drive/job/Drive/job/Drive-packages/1526/). Good to review @rom1win.

PR #1564, opened 2020-04-07
[WIP] NXDRIVE-2071: Move functional tests to GH Actions

BoboTiG:


BoboTiG:
So ...
Tests took 5-6h. They even timed out on Windows. This is due to the poor VM power (and only 2 CPUs). This is even more fragile: we did not use a local server, this would have been worst!

PR #1565, opened 2020-04-07
NXDRIVE-2023: Make the error title an URL to the remote document

rom1win:
Being able to click on the error title to open the document is an
intuitive action users often try.
The error title is now a link to the remote document.

### DIFF
--- a/nxdrive/data/qml/FileCard.qml
+++ b/nxdrive/data/qml/FileCard.qml
@@ -42,7 +42,7 @@ ShadowRectangle {
                 }
                 MouseArea {
                     anchors.fill: parent
-                    onClicked: api.open_local(engineUid, fileData.local_parent_path)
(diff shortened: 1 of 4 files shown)


nxdrive/data/qml/FileCard.qml
```diff
@@ -42,7 +42,7 @@ ShadowRectangle {
                 }
                 MouseArea {
                     anchors.fill: parent
-                    onClicked: api.open_local(engineUid, fileData.local_parent_path)
+                    onClicked: api.open_document(engineUid, type, fileData.remote_ref, fileData.remote_name, fileData.local_parent_path)
```
BoboTiG, line 45: This signature is too much. You should just pass `fileData`, it is a dict from `api._export_formatted_state(doc_pair)`.

PR #1593, opened 2020-04-20
NXDRIVE-2112: Always start a fresh download on Direct Edit

rom1win:
It seems that they are  more and more issues because of a previous
Direct Transfert download failed.
Doing another Direct Edit on the same document will resume the download.
Sometimes it will work, otherwise the download won't start.

We want to start a fresh download on each Direct Edit.


BoboTiG:
Clsing as it is not good for a review. It will trigger actions on each change else.

PR #1598, opened 2020-04-21
NXDRIVE-2112: Always start a fresh download on Direct Edit

rom1win:
It seems that there are more and more issues because of a previous
Direct Edit download failure. Doing another Direct Edit on the same
document will resume the download. Sometimes it will work, and
most of the time not.

A fresh download is now started everytime to prevent this issues.
The test test_resumed_download has benn added.

### DIFF
--- a/docs/changes/4.4.3.md
+++ b/docs/changes/4.4.3.md
@@ -5,6 +5,8 @@ Release date: `2020-xx-xx`
 ## Core
 
 - [NXDRIVE-1786](https://jira.nuxeo.com/browse/NXDRIVE-1786): Handle corrupted downloads in Direct Edit
+- [NXDRIVE-2112](https://jira.nuxeo.com/browse/NXDRIVE-2112): Always start a fresh download on Direct Edit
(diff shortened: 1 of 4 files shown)


docs/changes/4.4.3.md
```diff
@@ -5,6 +5,7 @@ Release date: `2020-xx-xx`
 ## Core
 
 - [NXDRIVE-1786](https://jira.nuxeo.com/browse/NXDRIVE-1786): Handle corrupted downloads in Direct Edit
+- [NXDRIVE-2112](https://jira.nuxeo.com/browse/NXDRIVE-2112): Always start a fresh download on Direct Edit
```
BoboTiG, line 8: Add the NXDRIVE-2116 too, it is the same fix and it is a good thing to reference it here.

nxdrive/direct_edit.py
```diff
@@ -290,6 +291,10 @@ def _download(
 
             pair = engine.dao.get_valid_duplicate_file(blob.digest)
 
+        # Remove output file if already exist. See NXDRIVE-2112
```
BoboTiG, line 294:
```suggestion
        # Remove the eventual temporary file. We do not want to be able to resume an
        # old download because of several issues and does not make sens for that feature.
        # See NXDRIVE-2112 and NXDRIVE-2116 for more context.
```

PR #1601, opened 2020-04-22
NXDRIVE-2113: [Direct Edit] Add a warning on upload when server-side lock is disabled

rom1win:
Uploading a new blob of an ongoing Direct Edit'ed document should be made only when the
user had already locked it.
The auto-locking of documents can be disabled on Drive throught an option, exposing
the user to possibles concurrent updates of the document.

A warning log has been added when a Direct Edit is done while this option is activated.


docs/changes/4.4.3.md
```diff
@@ -6,6 +6,7 @@ Release date: `2020-xx-xx`
 
 - [NXDRIVE-1786](https://jira.nuxeo.com/browse/NXDRIVE-1786): Handle corrupted downloads in Direct Edit
 - [NXDRIVE-2112](https://jira.nuxeo.com/browse/NXDRIVE-2112): Always start a fresh download on Direct Edit
+- [NXDRIVE-2113](https://jira.nuxeo.com/browse/NXDRIVE-2113): Always start a fresh download on Direct Edit
```
BoboTiG, line 9: Bad ticket.

PR #1605, opened 2020-04-23
NXDRIVE-2006: [Direct Transfer] Isolate the component

BoboTiG:
That refactoring was necessay:

- To be able to use the feature without synchronization.
- To be able to enable/disable the feature on-demand (modularity).
- The old code did not allow us to have global statistics
  (required by NXDRIVE-1926), it was too deeply inscruted in Drive.
  It was using signals and the sync engine to work.
  And the way it was done, there was no way to have more control
  over global metrics (plus several other smaller issues).
- To be able to setup multiple Direct Transfer session with priority.

* Introduced `state.py` to follow the application state from everywhere.

NXDRIVE-1910: Allow the feature when the application is paused

NXDRIVE-2029: Allow the feature when the synchronization is disabled

NXDRIVE-2018: Polish the component behavior


BoboTiG:
Will reopen.

PR #1616, opened 2020-04-28
NXDRIVE-2128: Allow only one Direct Edit'ion per document

rom1win:
When a document is opened, one should not be able to open it again.

### DIFF
--- a/nxdrive/direct_edit.py
+++ b/nxdrive/direct_edit.py
@@ -85,6 +85,7 @@ class DirectEdit(Worker):
         self.use_autolock = self._manager.get_direct_edit_auto_lock()
         self._event_handler: Optional[DriveFSEventHandler] = None
         self._metrics = {"edit_files": 0}
+        self._opened_files = set()
         self._observer: Observer = None
         self.local = LocalClient(self._folder)
         self._upload_queue: Queue = Queue()
@@ -531,7 +532,10 @@ class DirectEdit(Worker):
             self.directEditError.emit("DIRECT_EDIT_NOT_ENABLED", [])
             return
 
+        if doc_id in self._opened_files:
+            return
         log.info(f"Direct Editing doc {doc_id!r} on {server_url!r}")
+        self._opened_files.add(doc_id)
         try:
             # Download the file
             file_path = self._prepare_edit(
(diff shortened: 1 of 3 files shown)


tests/old_functional/test_direct_edit.py
```diff
@@ -486,6 +486,35 @@ def request(*args, **kwargs):
...
 
+    def test_multiple_open_file(self):
+        """Test to edit multiple times the same document while still opened.
+        """
+
+        def fake_open_local_file(*_, **__):
+            nonlocal called
+
+            called = True
+            open_local_file(*_, **__)
```
BoboTiG, line 497: You can remove that.

nxdrive/direct_edit.py
```diff
@@ -531,6 +532,8 @@ def edit(
             self.directEditError.emit("DIRECT_EDIT_NOT_ENABLED", [])
             return
 
+        if doc_id in self._opened_files and self._opend_files[doc_id]:
```
BoboTiG, line 535:
```suggestion
        if doc_id in self._opened_files and self._opened_files[doc_id]:
```

nxdrive/direct_edit.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, manager: "Manager", folder: Path) -> None:
         self.use_autolock = self._manager.get_direct_edit_auto_lock()
         self._event_handler: Optional[DriveFSEventHandler] = None
         self._metrics = {"edit_files": 0}
+        self._opened_files = {}
```
BoboTiG, line 88: You should use a `set`. Then, you can use `.discard()` and `.add()` in other places.

tests/old_functional/test_direct_edit.py
```diff
@@ -486,6 +486,32 @@ def request(*args, **kwargs):
...
+            nonlocal call_count
+            call_count += 1
+
+        with patch.object(self.manager_1, "open_local_file", new=open_local_file):
+            self.direct_edit.edit(*args, **kwargs)
+            assert self.direct_edit._opened_files == {uid: True}
+
+            self.direct_edit.edit(*args, **kwargs)
+            self.direct_edit.edit(*args, **kwargs)
+            self.direct_edit.edit(*args, **kwargs)
```
BoboTiG, line 512:
```suggestion
            self.direct_edit.edit(*args, **kwargs)
            assert self.direct_edit._opened_files == {uid: True}
```

nxdrive/direct_edit.py
```diff
@@ -657,6 +661,8 @@ def _handle_lock_queue(self) -> None:
                     continue
 
                 self.local.remove_remote_id(ref.parent, name="nxdirecteditlock")
+                if self._opened_files[ref]:
+                    self._opened_files.discard(ref)
```
BoboTiG, line 665: Just use `discard()`, no need for the previous check.

BoboTiG:
Ticket postponed as a real fix would be too random.

PR #1669, opened 2020-05-15
NXDRIVE-2129: [Direct Edit] Orphaned documents should first be remotely unlocked before being cleaned up

rom1win:
When the application is shut down during a direct edit, the edited file
stay locked on the remote server and the temporary file is not deleted.
When starting, the expected behaviour is to unlock the document on
the remote then remove the local orphan file.

### DIFF
--- a/tests/functional/test_direct_edit.py
+++ b/tests/functional/test_direct_edit.py
@@ -147,6 +147,69 @@ def test_cleanup_bad_folder_name(direct_edit):
     assert folder.is_dir()
 
 
+def test_cleanup_orphan_files(manager_factory, obj_factory):
+    """
+    Files marked as locked must be ignored by _cleanup().
+    This test goal is to ensure that these files are not removed at startup
+    as they are handled by the AutoLocker and the Direct Edit process later in the execution.
+    These files contains metadatas that are necessary for unlocking them on the server.
+    """
+    manager, engine = manager_factory()
+
+    with manager:
+        direct_edit = manager.direct_edit
+        direct_edit._folder.mkdir()
+        doc = obj_factory(title="this is a orphan file.odt", nature="File")
+
+        # Creating local orphan file
+        folder = direct_edit._folder / f"{doc.uid}_file-content"
+        folder.mkdir()
+        file = folder / "this is a orphan file.odt"
+        file.write_text("bla" * 3, encoding="utf-8")
+        assert file.is_file()
+
+        def get_locked_paths():
+            """
+            Mocked dao.get_local_paths() method.
+            Return the locked file path.
+            """
+            nonlocal file
+
+            return [file]
+
+        def extract_edit_info(_):
+            """
+            Mocked __extract_edit_info() method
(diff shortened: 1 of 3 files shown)


tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
     assert folder.is_dir()
 
 
+def test_cleanup_orphan_files(manager_factory, obj_factory):
+    """
+    Files marked as locked must be ignored by _cleanup.
```
BoboTiG, line 152:
```suggestion
    Files marked as locked must be ignored by _cleanup().
```

tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
...
+    These files are removed after there are unlocked.
```
BoboTiG, line 153:
```suggestion
    These files are removed after they are unlocked.
```

tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
...
+    """
+    manager, engine = manager_factory()
+
+    with manager:
+        direct_edit = manager.direct_edit
+        direct_edit._folder.mkdir()
+        doc = obj_factory(title="this is a orphan file.txt", nature="File")
```
BoboTiG, line 160: Just for the completeness of the test, do not use ".txt" as they will be seen as a Note. Use ".odt" instead:
```suggestion
        doc = obj_factory(title="this is a orphan file.odt", nature="File")
```

tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
...
+
+        # Creating local orphan file
+        folder = direct_edit._folder / f"{doc.uid}_file-content"
+        folder.mkdir()
+        file = folder / "this is a orphan file.txt"
```
BoboTiG, line 165: Same remark here.

tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
...
+        file.touch()
```
BoboTiG, line 166: Can be removed as the next line will create the file.

tests/functional/test_direct_edit.py
```diff
@@ -147,6 +147,68 @@ def test_cleanup_bad_folder_name(direct_edit):
...
+        file.write_text("bla" * 3, encoding="utf-8")
+        assert file.is_file()
+
+        def get_locked_paths():
+            """
+            Mocked dao.get_local_paths method.
```
BoboTiG, line 172:
```suggestion
            Mocked dao.get_local_paths() method.
```

PR #1727, opened 2020-06-08
NXDRIVE-2129: Fix behaviour of get_locked_paths()

rom1win:
When the application is shut down during a direct edit, the edited file
stay locked on the remote server and the temporary file is not deleted.
When starting, the expected behaviour is to unlock the document on
the remote then remove the local orphan file.

This commit fix the current behaviour in which Drive fails to retrieve
correctly the informations about the locked files from the database.
This prevent Drive to recognize orphan local files as locked.
This is the second step of a three steps fix of the bug.
Also changes has been updated.

### DIFF
--- a/tests/unit/test_autolock.py
+++ b/tests/unit/test_autolock.py
@@ -3,67 +3,60 @@
 Test the Auto-Lock feature used heavily by Direct Edit.
 """
 from pathlib import Path
-from typing import Dict, List, Tuple
+from typing import List, Tuple
 from unittest.mock import Mock, patch
 
 import nxdrive.autolocker
 import pytest
+from nxdrive.engine.dao.sqlite import ManagerDAO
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_autolock.py
```diff
@@ -3,80 +3,79 @@
...
 
     # Simulate watched files
-    autolock.set_autolock("already_locked.ods", Mock())
-    autolock.set_autolock("abc こん ツリー/2.ods", Mock())
+    file1 = autolock._folder / "already_locked.ods"
+    file1.touch()
+    autolock.set_autolock(file1, Mock())
+    autolock.set_autolock(autolock._folder / "abc こん ツリー.ods", Mock())
+
+    # Another file, noy yet watched
```
BoboTiG, line 42:
```suggestion
    # Another file, not yet watched
```

PR #1805, opened 2020-07-01
NXDRIVE-2193: Stop trying to guess the server URL

rom1win:
When adding an account the app was trying to understand
the given URL by deconstructing and reconstructing it using
different patterns. This behaviour leaded to several issues
and was a poor end user feature in the end.
From now, a simple syntax check will be done on the URL and if
it is wrong then the user will have to fix it.

Changelog has been updated.


tests/unit/test_utils.py
```diff
@@ -684,56 +683,6 @@ def test_get_value(raw_value, expected_value):
...
-
-    for generated_url in nxdrive.utils.compute_urls(url):
-        # There should be only one "//" in each URL
-        assert generated_url.count("//") == 1
-
-        # There should be no whitespace
-        assert not no_whitespace.findall(generated_url)
-
-
-def test_guess_server_url_bad_ssl():
```
BoboTiG, line 721: We need to keep this test.

BoboTiG:
Could you rebase the PR on master please?

docs/changes/4.4.5.md
```diff
@@ -43,4 +43,6 @@ Release date: `2020-xx-xx`
 
 ## Technical Changes
 
--
+- Removed utils.py::`guess_server_url()`
+- Removed utils.py::`compute_urls()`
```
BoboTiG, line 47: Reverse those 2 lines (alpha sort).

PR #1808, opened 2020-07-02
NXDRIVE-2219: Display a remote folder icon for each transfer

BoboTiG:
This fixes UI performance issues because there is no need to do QML > C > Python > C > QML to fetch the remote folder path and UID (it was displayed in the header of the Direct transfer window).

Another performance issue was tackled: the use of xattrs to store remote paths and UIDs. Sending 50,000 files was freezing the UI; not it is still frozen but quite less. I may need additionnal work to completely tackle that issue.

Some interesting fixes/improvements:

- A nice folder icon is shown on the left of each progress bar to access the remote destination folder.
- Clean-up Direct transfer tests.
- Notifications are shown when the uploaded file is >= 25 MiB.
- The buttons available when cancelling a transfer are now Yes/No.


BoboTiG:
Still need to find out to prevent the window being closed when there are still transfers to handle but there is a small time they are not available.

PR #1812, opened 2020-07-03
NXDRIVE-2221: Improve performances to prevent GUI freezes

BoboTiG:
- Do less work on the QML side.
- Improved upload file size computation: it is now done one-time only.
- Fixed duplicate prevention for documents without blobs.
- File size is only computed once for all `FileAction`s.
- Fixed `KeyError` in `BaseUploader.upload_chunks()` (`upload_idx`).
- A small sync icon + count is displayed at the bottom right.
- Items to transfer are first added to the database and then started. This is proper and will prevent GUI freeze because of too many bidirectional database access.

See my comment on https://jira.nuxeo.com/browse/NXDRIVE-2221 for more details + perf reports.

### DIFF
--- a/nxdrive/engine/dao/sqlite.py
+++ b/nxdrive/engine/dao/sqlite.py
@@ -23,6 +23,7 @@ from threading import RLock, local
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     Generator,
     List,
@@ -619,6 +620,7 @@ class EngineDAO(ConfigurationDAO):
 
     newConflict = pyqtSignal(object)
     transferUpdated = pyqtSignal()
+    directTransferUpdated = pyqtSignal()
 
     def __init__(self, db: Path) -> None:
         super().__init__(db)
@@ -630,7 +632,7 @@ class EngineDAO(ConfigurationDAO):
         self.reinit_processors()
 
     def get_schema_version(self) -> int:
-        return 12
+        return 13
 
     def _migrate_state(self, cursor: Cursor) -> None:
         try:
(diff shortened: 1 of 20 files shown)


nxdrive/gui/api.py
```diff
@@ -183,43 +184,53 @@ def app_update(self, version: str) -> None:
...
+        for count, upload in enumerate(dao.get_uploads()):
+            if count >= limit:
+                break
+            result.append(asdict(upload))
 
         return result
 
-    @pyqtSlot(str, str, int, float)
+    def get_direct_transfer_items(self, dao: EngineDAO) -> List[Dict[str, Any]]:
+        return list(dao.get_dt_uploads_raw())
```
BoboTiG, line 204: Limit to 5 items.

nxdrive/engine/dao/sqlite.py
```diff
@@ -2175,22 +2185,61 @@ def get_uploads(self) -> Generator[Upload, None, None]:
...
+            yield Upload(
+                res.uid,
+                Path(res.path),
+                TransferStatus(res.status),
+                res.engine,
+                doc_pair=res.doc_pair,
+                batch=json.loads(res.batch),
+            )
+
+    def get_dt_uploads_raw(self) -> Generator[Dict[str, Any], None, None]:
```
BoboTiG, line 2210: Check if this is still needed. Now that lot of fixes were done, it may be useless to have another method such this one.

nxdrive/engine/engine.py
```diff
@@ -578,32 +606,74 @@ def resume(self) -> None:
...
+        """
+        Resume downloads having the *status* status.
+        Return the list of doc_pair UIDs that were resumed.
+        """
+        for transfer in self.dao.get_downloads_with_status(status):
+            if transfer.uid is not None:
+                self.dao.resume_transfer("download", transfer.uid)
+                yield transfer.doc_pair
+
+    def _resume_uploads(
```
BoboTiG, line 641: Do a better refactorinf of those methods. Too repetitives.

BoboTiG:
Note for myself: we should wait for all database insertions before starting to send files.

PR #1815, opened 2020-07-06
NXDRIVE-2226: Disable temporarily S3 capability

rom1win:
Due to NXDRIVE-2210, the S3 feature is now temporarily disabled.

Changelog has been updated.


nxdrive/data/qml/FeaturesTab.qml
```diff
@@ -41,6 +41,7 @@ Rectangle {
 
                         text: item[0] + (beta_features.includes(item[1]) ? sup_tag: "")
                         checked: manager.get_feature_state(item[1])
+                        enabled: disabled_features.includes(item[1]) ? false: true
```
BoboTiG, line 44: It is enough to do?
```suggestion
                        enabled: disabled_features.includes(item[1])
```

PR #1850, opened 2020-07-15
NXDRIVE-2186: Do not remove staled transfers at startup if the applic…

BoboTiG:
…ation crashed


BoboTiG:
GH does not seem to be good for Actions. All tests pass locally, I am merging then.

PR #1859, opened 2020-07-17
NXDRIVE-2144: [Direct Edit] Prevent unintentional upload of files

rom1win:
When Direct Edit'ing a file, a temporary file may be created next to
the edited file. During Drive startup, all the orphaned Direct Edit'ed files
are listed. If a temporary file is listed before the orphaned file then
it will be uploaded instead of the original file.

Some security has been added to prevent Drive from uploading these files
as it may result in data loss on Nuxeo.

### DIFF
--- a/tests/functional/test_direct_edit.py
+++ b/tests/functional/test_direct_edit.py
@@ -8,6 +8,7 @@ import pytest
 from nuxeo.exceptions import CorruptedFile
 from nxdrive.constants import ROOT
 from nxdrive.engine.engine import Engine, ServerBindingSettings
+from nxdrive.objects import DirectEditDetails
 from nxdrive.translator import Translator
 from nxdrive.utils import find_resource
 
@@ -179,13 +180,9 @@ def test_cleanup_orphan_files(manager_factory, obj_factory):
 
         def extract_edit_info(_):
             """
-            Mocked __extract_edit_info() method
-            Mandatory as there is no blob to work with in the test
+            Mocked _extract_edit_info() method.
+            Mandatory as there is no blob to work with in the test.
             """
-            DirectEditDetails = namedtuple(
-                "details",
-                ["uid", "engine", "digest_func", "digest", "xpath", "editing"],
-            )
 
             nonlocal doc
             nonlocal engine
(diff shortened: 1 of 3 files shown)


tests/functional/test_direct_edit.py
```diff
@@ -210,6 +210,77 @@ def extract_edit_info(_):
         assert file.is_file()
 
 
+def test_office_temporary_files(manager_factory, obj_factory):
```
BoboTiG, line 213:
```suggestion
def test_document_hijacking(manager_factory, obj_factory):
```

tests/functional/test_direct_edit.py
```diff
@@ -210,6 +210,77 @@ def extract_edit_info(_):
...
+    """
+    Software from the office suite may leave temporary file behind them.
+    This test goal is to ensure that these files are not removed at startup
+    nor uploaded to the remote by the _cleanup method.
+    Uploading a temporary file may result in data loss on the remote.
```
BoboTiG, line 218:
```suggestion
    There are 2 situations we need to prevent:

    1. Softwares LibreOffice or MS Office may leave temporary files behind them.
    2. Someone could intentionally put a bad file in the good folder.

    Both will corrupt the remote document (data loss) at the next startup.

    The test goal is to ensure that these files are not just ignored.
```

tests/functional/test_direct_edit.py
```diff
@@ -210,6 +210,77 @@ def extract_edit_info(_):
...
+
+        # Creating Excel extensionless temp file
+        random_file = folder / "532AEF06"
+        random_file.write_text("bla" * 3, encoding="utf-8")
+        assert random_file.is_file()
+
+        def extract_edit_info(ref):
+            """
+            Mocked __extract_edit_info() method
+            Mandatory as there is no blob to work with in the test
```
BoboTiG, line 252:
```suggestion
            Mocked _extract_edit_info() method.
            Mandatory as there is no blob to work with in the test.
```

tests/functional/test_direct_edit.py
```diff
@@ -210,6 +210,77 @@ def extract_edit_info(_):
...
+            """
+            DirectEditDetails = namedtuple(
```
BoboTiG, line 254: Use the one from `objects.py`.

tests/functional/test_direct_edit.py
```diff
@@ -210,6 +210,77 @@ def extract_edit_info(_):
...
+        with patch.object(
+            direct_edit, "_extract_edit_info", new=extract_edit_info
+        ), ensure_no_exception():
+            direct_edit._cleanup()
+
+        # The files should still be present as they are not deleted by the clean-up
+        assert xlsx_file.is_file()
+        assert tmp_file.is_file()
+        assert random_file.is_file()
+
```
BoboTiG, line 282: Also check the upload queue or whatever object to ensure there will be nothing uploaded.

PR #1881, opened 2020-07-24
NXDRIVE-2254: Always send primary metrics (app and OS versions)

BoboTiG:
Send a simple "Hello world" event at startup.


BoboTiG:
Test failures are related to https://community.nuxeo.com that is currently down.

BoboTiG:
Test failures are related to https://community.nuxeo.com that is currently down.

PR #1916, opened 2020-08-05
NXDRIVE-2266: [Direct Transfer] Should ignore same file patterns as the sync engine

rom1win:
Hidden files are now ignored during Direct Transfer.

### DIFF
--- a/nxdrive/engine/engine.py
+++ b/nxdrive/engine/engine.py
@@ -430,12 +430,20 @@ class Engine(QObject):
         items = []
 
         for local_path in sorted(local_paths):
+            name = local_path.name
+            if name.startswith(Options.ignored_prefixes) or name.endswith(
+                Options.ignored_suffixes
+            ):
+                log.debug(
+                    f"Cannot add {local_path!r} to Direct Transfer queue as it is ignored."
+                )
+                continue
             if local_path.is_file():
                 items.append(
                     (
                         str(local_path),
                         str(local_path.parent),
-                        local_path.name,
+                        name,
                         False,
                         local_path.stat().st_size,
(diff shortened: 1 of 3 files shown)


nxdrive/engine/engine.py
```diff
@@ -430,12 +430,17 @@ def _direct_transfer(
         items = []
 
         for local_path in sorted(local_paths):
+            name = local_path.name
             if local_path.is_file():
+                if name.startswith(Options.ignored_prefixes) or name.endswith(
+                    Options.ignored_suffixes
+                ):
+                    continue
```
BoboTiG, line 438: Add a debug log to help following what happens.

nxdrive/engine/engine.py
```diff
@@ -447,13 +452,18 @@ def _direct_transfer(
             else:
                 tree = sorted(get_tree_list(local_path, remote_parent_path))
                 for path, remote_subparent_path, size in tree:
+                    name = path.name
+                    if name.startswith(Options.ignored_prefixes) or name.endswith(
+                        Options.ignored_suffixes
+                    ):
+                        continue
```
BoboTiG, line 459: Add a debug log to help following what happens.

nxdrive/engine/engine.py
```diff
@@ -430,12 +430,20 @@ def _direct_transfer(
         items = []
 
         for local_path in sorted(local_paths):
+            name = local_path.name
             if local_path.is_file():
+                if name.startswith(Options.ignored_prefixes) or name.endswith(
```
BoboTiG, line 435: Move this line before the `if local_path.is_file():`.

PR #1920, opened 2020-08-05
NXDRIVE-2147: [Direct Edit] Add a maximum retry count for items in the upload queue

rom1win:
Unhandled error while uploading a file where retried indefinitely.
This was quite bad as it exploded server logs, client logs and
the Sentry quote. From now, a failing upload will be tried 3 times
before being dropped from the upload_queue.

### DIFF
--- a/nxdrive/direct_edit.py
+++ b/nxdrive/direct_edit.py
@@ -2,6 +2,7 @@
 import errno
 import re
 import shutil
+from collections import defaultdict
 from contextlib import suppress
 from datetime import datetime
 from logging import getLogger
@@ -88,8 +89,9 @@ class DirectEdit(Worker):
         self._observer: Observer = None
         self.local = LocalClient(self._folder)
         self._upload_queue: Queue = Queue()
+        self._upload_errors: Dict[Path, int] = defaultdict(int)
         self._lock_queue: Queue = Queue()
-        self._error_queue = BlacklistQueue()
+        self._error_queue = BlacklistQueue(delay=Options.delay)
         self._stop = False
         self.watchdog_queue: Queue = Queue()
 
@@ -717,7 +719,7 @@ class DirectEdit(Worker):
     def _handle_upload_queue(self) -> None:
         while "items":
             try:
-                ref = self._upload_queue.get_nowait()
+                ref: Path = self._upload_queue.get_nowait()
             except Empty:
                 break
 
@@ -824,7 +826,7 @@ class DirectEdit(Worker):
             except CONNECTION_ERROR:
                 # Try again in 30s
                 log.warning(f"Connection error while uploading {ref!r}", exc_info=True)
-                self._error_queue.push(ref)
+                self._handle_upload_error(ref, os_path)
             except HTTPError as e:
                 if e.status == 500 and "Cannot set property on a version" in e.message:
                     log.warning(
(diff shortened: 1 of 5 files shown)


nxdrive/direct_edit.py
```diff
@@ -63,6 +63,7 @@ def _is_lock_file(name: str) -> bool:
 class DirectEdit(Worker):
     localScanFinished = pyqtSignal()
     directEditUploadCompleted = pyqtSignal(str)
+    directEditUploadFailed = pyqtSignal(str)
```
BoboTiG, line 66: I would prefer to reuse `directEditError` as we are doing at several places. WDYT?

nxdrive/direct_edit.py
```diff
@@ -869,6 +891,7 @@ def _execute(self) -> None:
         try:
             self._cleanup()
             self._setup_watchdog()
+            print("here")
```
BoboTiG, line 894: To remove.

nxdrive/notification.py
```diff
@@ -421,6 +421,23 @@ def __init__(self, filename: str) -> None:
         )
 
 
+class DirectEditUploadFailedNotification(Notification):
```
BoboTiG, line 424: To remove if we go with the `directEditError` signal.

nxdrive/notification.py
```diff
@@ -601,6 +618,9 @@ def _directEditStarting(self, hostname: str, filename: str) -> None:
     def _directEditUpdated(self, filename: str) -> None:
         self.send_notification(DirectEditUpdatedNotification(filename))
 
+    def _directEditUploadFailed(self, filename: str) -> None:
```
BoboTiG, line 621: To remove if we go with the `directEditError` signal.

tests/old_functional/test_direct_edit.py
```diff
@@ -319,6 +319,40 @@ def upload(*_, **__):
         # And the error queue should still be be empty, the bad upload has beend discarded
         assert self.direct_edit._error_queue.empty()
 
+    def test_direct_edit_max_error(self):
+        """
+        When uploading changes to the remote, some error can be raised.
```
BoboTiG, line 324:
```suggestion
        When uploading changes to the server, recurrent errors can happen.
```

nxdrive/direct_edit.py
```diff
@@ -839,11 +844,28 @@ def _handle_upload_queue(self) -> None:
...
+            del self._upload_errors_watcher[key]
+            return
+        elif (
+            key in self._upload_errors_watcher and self._upload_errors_watcher[key] < 3
+        ):
+            self._upload_errors_watcher[key] += 1
+        else:
+            self._upload_errors_watcher[key] = 1
+        self._error_queue.push(ref)
+        print(self._upload_errors_watcher)
```
BoboTiG, line 868: Hm this code is not beautful to read. Maybe could we leverage [collections.defaultdict()](https://docs.python.org/3/library/collections.html#collections.defaultdict)?

Something like the following.

In `__init__()`, change to:
```python
    from collections import defaultdict  # move to the top, of course
    self._upload_errors_watcher: Dict[str, int] = defaultdict(int)
```

And adapt that method to something like:
```suggestion
    def _handle_upload_error(self, ref: Path) -> None:
        # (Add the missing docstring)
        self._upload_errors_watcher[ref] += 1
        if self._upload_errors_watcher[key] < 3:
            self._error_queue.push(ref)
            return

        log.error(f"Upload queue error: too many failures for {ref!r}, skipping the upload")
        self.directEditUploadFailed.emit(ref.name)  # TODO: adapt with directEditError
        self._upload_errors_watcher.pop(key, None)

```

Note: I did not test the code.

tests/old_functional/test_direct_edit.py
```diff
@@ -319,6 +319,40 @@ def upload(*_, **__):
...
+            self.wait_sync(timeout=120, fail_if_timeout=False)
+
+            # The file should _not_ be updated on the server
+            assert (
+                self.remote.get_blob(self.remote.get_info(doc_id))
+                == b"Initial content."
+            )
+
+        # There must be no error has the upload has been dropped.
+        assert self.direct_edit._error_queue.empty()
```
BoboTiG, line 354: Also check for the `_upload_error_queue`.

nxdrive/direct_edit.py
```diff
@@ -839,11 +841,21 @@ def _handle_upload_queue(self) -> None:
...
-                    self._error_queue.push(ref)
+                    self._handle_upload_error(ref)
             except Exception:
                 # Try again in 30s
                 log.exception(f"Direct Edit unhandled error for ref {ref!r}")
-                self._error_queue.push(ref)
+                self._handle_upload_error(ref)
+
+    def _handle_upload_error(self, ref: Path, max_count: int = 4) -> None:
+        """Update the _upload_errors dict and handle the error accordingly to the max_count."""
```
BoboTiG, line 851:
```suggestion
        """Retry the upload if the number of attempts is below *max_count* else discard it."""
```

PR #1932, opened 2020-08-12
NXDRIVE-2277: [Direct Edit] Changes not uploaded anymore when saving file on Windows

rom1win:
The error was due to a bad comparison in `utils.py::normalize_event_filename()`. 
The method has been fixed and now use windows long path.


BoboTiG:
If I am right, the regression was introduced by 41f37bed79254c832f805e5aa5b68210a4a23c9a. I would prefer to revert that commit instead of altering `normalize_event_filename()` which is a sensitive function called a lot of times; and I am not sure of all possible side effects that would imply.

BoboTiG:
The night has worked :)
I will merge only the specific code that applies to the patch (not the test). Let's see with the beta if something araises.

PR #2024, opened 2020-09-17
NXDRIVE-2076: Change the cursor on click on new account connect button

rom1win:
When adding an account, the browser may take some time
to open. A waiting cursor has been added so the user
can know that something is currently happening.

### DIFF
--- a/nxdrive/gui/application.py
+++ b/nxdrive/gui/application.py
@@ -2,6 +2,7 @@
 """ Main Qt application handling OS events and system tray UI. """
 import os
 import sys
+import webbrowser
 from functools import partial
 from logging import getLogger
 from math import sqrt
@@ -849,7 +850,11 @@ class Application(QApplication):
             through the app, it opens in the browser and retrieves the login token
             by opening an nxdrive:// URL.
             """
-            self.manager.open_local_file(url)
+            QApplication.setOverrideCursor(Qt.WaitCursor)
+            try:
+                webbrowser.open_new_tab(url)
+            finally:
+                QApplication.restoreOverrideCursor()
         else:
             self._web_auth_not_frozen(url)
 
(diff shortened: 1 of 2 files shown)


nxdrive/gui/application.py
```diff
@@ -839,7 +839,11 @@ def confirm_cancel_transfer(
     def open_authentication_dialog(
         self, url: str, callback_params: Dict[str, str]
     ) -> None:
+        from PyQt5.QtCore import Qt
+        from PyQt5.QtWidgets import QApplication
+
         self.api.callback_params = callback_params
+        QApplication.setOverrideCursor(Qt.WaitCursor)
```
BoboTiG, line 846: - Only change the cursor when the app is frozen.
- Use a try/finally block to ensure the cursor is always restored.

BoboTiG (changes requested):

BoboTiG (approved):

nxdrive/gui/application.py
```diff
@@ -849,7 +850,11 @@ def open_authentication_dialog(
             through the app, it opens in the browser and retrieves the login token
             by opening an nxdrive:// URL.
             """
-            self.manager.open_local_file(url)
+            try:
+                QApplication.setOverrideCursor(Qt.WaitCursor)
```
BoboTiG, line 854: Move this call just before the "try".

nxdrive/gui/application.py
```diff
@@ -849,7 +850,11 @@ def open_authentication_dialog(
...
+                webbrowser.open_new_tab(url)
```
BoboTiG, line 855: 👍

BoboTiG (changes requested):

BoboTiG (approved):

PR #2028, opened 2020-09-18
NXDRIVE-2306: Do not validate again settings when the value did not change

rom1win:
When validating a setting, a QML check is done to check if
the setting value has changed and prevent useless calls to
Python code.


nxdrive/data/qml/ProxyPopup.qml
```diff
@@ -112,8 +112,14 @@ NuxeoPopup {
                     || (isAuto && pacUrlInput.text)
                 inverted: true
                 onClicked: {
+                    var current_proxy_settings = JSON.parse(api.get_proxy_settings())
```
BoboTiG, line 115: Did you try that change?

BoboTiG (approved):

PR #2111, opened 2020-10-16
NXDRIVE-2309: Update view for migrated sessions

rom1win:
When the completed_on infromation is missing for a previsously
created sessions, a simple Cancelled or Completed is shown
instead of the '?'.


nxdrive/gui/view.py
```diff
@@ -394,29 +396,42 @@ def data(self, index: QModelIndex, role: int = REMOTE_PATH) -> Any:
...
+            if datetime:
+                label += "_ON"
+                offset = tzlocal().utcoffset(datetime)
+                if offset:
+                    datetime += offset
+                args.append(Translator.format_datetime(datetime))
+            return self.tr(label, args)
+        elif role == self.PROGRESS:
+            label = "SESSION_PROGRESS"
+            args = [f"{row['uploaded']}/{row['total']}"]
```
BoboTiG, line 433: And that's even better because now we can use number formatting :)
Look how `sizeof_fmt()` is used in that file to do the same here.
Maybe adding spaces around the slash would be a good thing too?

nxdrive/gui/view.py
```diff
@@ -366,9 +367,10 @@ def __init__(self, translate: Callable, parent: QObject = None) -> None:
             self.UPLOADED: b"uploaded",
             self.TOTAL: b"total",
             self.ENGINE: b"engine",
-            self.CREATED_AT: b"created_at",
-            self.COMPLETED_AT: b"completed_at",
+            self.CREATED_ON: b"created_on",
+            self.COMPLETED_ON: b"completed_on",
```
BoboTiG, line 371: You have to update `Session` attributes also to keep the code coherent and easy reading/searching the whole code base.

nxdrive/gui/view.py
BoboTiG: OK I mixed my thoughts. You can skip that review.

BoboTiG: But still, you have to use a delimiter.

nxdrive/gui/view.py
```diff
@@ -394,29 +396,42 @@ def data(self, index: QModelIndex, role: int = REMOTE_PATH) -> Any:
...
+            if datetime:
+                label += "_ON"
+                offset = tzlocal().utcoffset(datetime)
+                if offset:
+                    datetime += offset
+                args.append(Translator.format_datetime(datetime))
+            return self.tr(label, args)
+        elif role == self.PROGRESS:
+            label = "SESSION_PROGRESS"
+            args = [f"{row['uploaded']} / {row['total']}"]
```
BoboTiG, line 433:
```suggestion
            args = [f"{row['uploaded']:,} / {row['total']:,}"]
```

nxdrive/gui/view.py
```diff
@@ -485,29 +502,42 @@ def data(self, index: QModelIndex, role: int = REMOTE_PATH) -> Any:
...
+            if datetime:
+                label += "_ON"
+                offset = tzlocal().utcoffset(datetime)
+                if offset:
+                    datetime += offset
+                args.append(Translator.format_datetime(datetime))
+            return self.tr(label, args)
+        elif role == self.PROGRESS:
+            label = "SESSION_PROGRESS"
+            args = [f"{row['uploaded']}"]
```
BoboTiG, line 539:
```suggestion
            args = [f"{row['uploaded']:,}"]
```

BoboTiG (changes requested):

BoboTiG (approved):

PR #2151, opened 2020-10-28
NXDRIVE-2309: [Direct Transfer] Optimize resume_session() and cancel_session() SQL requests

rom1win:


nxdrive/engine/dao/sqlite.py
```diff
@@ -2735,6 +2735,24 @@ def resume_transfer(
...
+            )
+            rows = c.execute(
+                "SELECT * FROM States WHERE id IN"
+                " (SELECT doc_pair FROM Uploads WHERE doc_pair IN (SELECT id FROM States WHERE Session = ?))",
+                (uid,),
+            ).fetchall()
+
+            for doc_pair in rows:
+                self.queue_manager.push(doc_pair)
+                self.directTransferUpdated.emit()
```
BoboTiG, line 2754: To be even more efficient, should that line be unindented? It would trigger the signal only once, after all pairs are in the queue.

nxdrive/engine/dao/sqlite.py
```diff
@@ -2746,6 +2765,33 @@ def pause_session(self, uid: int) -> None:
...
+                    "SELECT * FROM Uploads WHERE doc_pair IN (SELECT id FROM States WHERE Session = ?)",
+                    (uid,),
+                ).fetchall()
+            ]
+            c.execute(
+                "DELETE FROM Uploads WHERE doc_pair IN (SELECT id FROM States WHERE Session = ?)",
+                (uid,),
+            )
+            c.execute("DELETE FROM States WHERE Session = ?", (uid,))
+            self.directTransferUpdated.emit()
```
BoboTiG, line 2784: Why not putting that at the end of the method?

tests/old_functional/test_direct_transfer.py
```diff
@@ -986,7 +986,7 @@ def callback(*_):
             # A new Notification logs should appear at each iteration
             records = map(str, self._caplog.records)
             matches = list(filter(expression.match, records))
-            assert len(matches) == 1
+            assert len(matches) == 0
```
BoboTiG, line 989: Why not removing the whole notification check from the test?

nxdrive/engine/engine.py
```diff
@@ -713,12 +711,10 @@ def cancel_upload(self, transfer_uid: int) -> None:
...
         """Cancel all transfers for given session."""
-        session_uploads = self.dao.get_session_uploads(uid)
-        if not session_uploads:
-            return
-        for upload in session_uploads:
-            self.cancel_upload(upload["uid"])
         self.dao.change_session_status(uid, TransferStatus.CANCELLED)
+        batchs_to_cancel = self.dao.cancel_session(uid)
+        for batch in batchs_to_cancel:
+            self.remote.cancel_batch(batch)
```
BoboTiG, line 717:
```suggestion
        self.dao.cancel_session(uid)
        # We could cancel all batches, but in reality it would freeze the GUI.
        # Cancelling a session with a lot of items is worthless the use of a specific thread
        # to do the clean-up ourselves. Let the server doing it for us. We will be able to
        # tackle a possible issue in time.
        # for batch in self.dao.cancel_session(uid):
        #     self.remote.cancel_batch(batch)
```

nxdrive/engine/dao/sqlite.py
```diff
@@ -2735,17 +2719,64 @@ def resume_transfer(
...
+            self.change_session_status(uid, TransferStatus.PAUSED)
             c.execute(
                 "UPDATE Uploads SET status = ? WHERE doc_pair IN (SELECT id FROM States WHERE Session = ?)",
                 (TransferStatus.PAUSED.value, uid),
             )
-            self.change_session_status(uid, TransferStatus.PAUSED)
             self.directTransferUpdated.emit()
 
+    def cancel_session(self, uid: int) -> List[Dict[str, Any]]:
+        """Cancel all transfers for given session."""
```
BoboTiG, line 2753:
```suggestion
        """Cancel all transfers for given session.
        Return the list of impacted batches to be able to cancel them later.
        """
```

BoboTiG (approved):

PR #2160, opened 2020-11-02
NXDRIVE-2309: [Direct Transfer] Fix multiple non-chunked uploads pause and cancel

rom1win:
Direct Transfer uploads will now inherit the session status when saved into database. Uploads from cancelled sessions won't be uploaded anymore.

### DIFF
--- a/nxdrive/engine/dao/sqlite.py
+++ b/nxdrive/engine/dao/sqlite.py
@@ -2667,6 +2667,47 @@ class EngineDAO(ConfigurationDAO):
             else:
                 self.transferUpdated.emit()
 
+    def save_dt_upload(self, upload: Upload) -> None:
+        """
+        New Direct Transfer upload.
+        Will have the same status as it's session.
+        """
+        with self.lock:
+            # Remove non-serializable data, never used elsewhere
+            batch = {k: v for k, v in upload.batch.items() if k != "blobs"}
+
+            values = (
+                upload.path,
+                upload.doc_pair,
+                # Default value if IFNULL is validated, meaning that the linked state has been removed.
+                TransferStatus.CANCELLED.value,
+                upload.engine,
+                upload.is_direct_edit,
+                upload.is_direct_transfer,
+                upload.filesize,
+                json.dumps(batch),
+                upload.chunk_size,
+                upload.remote_parent_path,
+                upload.remote_parent_ref,
+                upload.doc_pair,
+            )
+            c = self._get_write_connection().cursor()
+            sql = (
+                "INSERT INTO Uploads "
+                "(path, status, engine, is_direct_edit, is_direct_transfer, filesize, batch, chunk_size,"
+                " remote_parent_path, remote_parent_ref, doc_pair)"
+                " VALUES (?, IFNULL((SELECT s.status FROM States st INNER JOIN Sessions s ON st.session = s.uid "
+                "AND st.id = ?), ?), ?, ?, ?, ?, ?, ?, ?, ?, ?)"
+            )
+            c.execute(sql, values)
+
+            # Important: update the upload UID and status attr
+            upload.uid = int(c.execute("SELECT last_insert_rowid()").fetchone()[0])
+            res = self.get_dt_upload(uid=upload.uid)
+            # Upload may be deleted right after creation by session cancel.
+            upload.status = res.status if res else TransferStatus.CANCELLED
+            self.directTransferUpdated.emit()
+
     def update_upload(self, upload: Upload) -> None:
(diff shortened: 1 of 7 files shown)


nxdrive/engine/dao/sqlite.py
```diff
@@ -2660,6 +2660,45 @@ def save_upload(self, upload: Upload) -> None:
             else:
                 self.transferUpdated.emit()
 
+    def save_dt_upload(self, upload: Upload) -> None:
+        """
+        New direct transfer upload.
+        Woll have the same status as it's Session on creation.
```
BoboTiG, line 2666:
```suggestion
        New Direct Transfer upload.
        Will have the same status as it's session.
```

nxdrive/engine/dao/sqlite.py
```diff
@@ -2660,6 +2660,45 @@ def save_upload(self, upload: Upload) -> None:
...
+        """
+        with self.lock:
+            # Remove non-serializable data, never used elsewhere
+            batch = {k: v for k, v in upload.batch.items() if k != "blobs"}
+
+            values = (
+                upload.path,
+                upload.doc_pair,
+                TransferStatus.CANCELLED.value,
```
BoboTiG, line 2675: This should be the session status, right?

nxdrive/engine/dao/sqlite.py
```diff
@@ -2660,6 +2660,45 @@ def save_upload(self, upload: Upload) -> None:
...
+                upload.remote_parent_ref,
+                upload.doc_pair,
+            )
+            c = self._get_write_connection().cursor()
+            sql = (
+                "INSERT INTO Uploads "
+                "(path, status, engine, is_direct_edit, is_direct_transfer, filesize, batch, chunk_size,"
+                " remote_parent_path, remote_parent_ref, doc_pair)"
+                " VALUES (?, IFNULL((SELECT s.status FROM States INNER JOIN Sessions s ON States.session = s.uid "
+                "AND States.id = ?), ?), ?, ?, ?, ?, ?, ?, ?, ?, ?)"
```
BoboTiG, line 2692:
```suggestion
                " VALUES (?, IFNULL((SELECT s.status FROM States st INNER JOIN Sessions s ON st.session = s.uid "
                "AND st.id = ?), ?), ?, ?, ?, ?, ?, ?, ?, ?, ?)"
```

nxdrive/engine/dao/sqlite.py
```diff
@@ -2660,6 +2660,45 @@ def save_upload(self, upload: Upload) -> None:
...
+            )
+            c.execute(sql, values)
+
+            # Important: update the upload UID and status attr
+            upload.uid = int(c.execute("SELECT last_insert_rowid()").fetchone()[0])
+            res = self.get_dt_upload(uid=upload.uid)
+            upload.status = res.status if res else TransferStatus.CANCELLED
```
BoboTiG, line 2699: I do not like that idea. Could we just passe the sessionas argument to prevent requerying the database?

nxdrive/exceptions.py
```diff
@@ -158,6 +158,12 @@ class UploadPaused(TransferPaused):
     pass
 
 
+class UploadCancelled(DriveError):
```
BoboTiG, line 161: You should crzate a parent class `TransferCancelled` and inherit from it. Like it is done for  `UploadPaused`.

nxdrive/engine/processor.py
```diff
@@ -367,6 +368,8 @@ def _execute(self) -> None:
                 self.engine.dao.set_transfer_doc(
                     nature, exc.transfer_id, self.engine.uid, doc_pair.id
                 )
+            except UploadCancelled:
+                pass
```
BoboTiG, line 372:
```suggestion
            except UploadCancelled as exc:
                log.debug(f"Cancelled upload {exc.transfer_id!r}")
```

nxdrive/engine/dao/sqlite.py
BoboTiG: Add a comment to ease understanding, this is way too magic combined with SQL `IFNULL()`.

nxdrive/exceptions.py
```diff
@@ -146,6 +146,13 @@ def __init__(self, transfer_id: int) -> None:
         self.transfer_id = transfer_id
 
 
+class TransferCancelled(DriveError):
```
BoboTiG, line 149: Move it above `TransferPaused`, keep it sorted.

BoboTiG (commented):
Fix tests, feed the changelog and we are good :)

PR #2265, opened 2020-12-01
NXDRIVE-2432: [Direct Transfer] Only display chunked transfers into the Monitoring tab

rom1win:
Only chunked files are now displayed in the Direct Transfer Monitoring tab.

### DIFF
--- a/nxdrive/data/qml/DirectTransfer.qml
+++ b/nxdrive/data/qml/DirectTransfer.qml
@@ -9,36 +9,20 @@ Rectangle {
     anchors.fill: parent
 
     property string engineUid: ""
-    property int itemsCount: 0
     property int activeSessionsCount: 0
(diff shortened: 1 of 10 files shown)


nxdrive/data/qml/DirectTransfer.qml
```diff
@@ -160,48 +144,34 @@ Rectangle {
...
 
-                    model: DirectTransferModel
-                    delegate: TransferItem {}
-                }
+                Layout.fillWidth: true
+                Layout.fillHeight: true
+                ScrollBar.vertical: ScrollBar {}
             }
 
             // The number of total items to sync, including ones being synced
```
BoboTiG, line 166: Need to update that comment too.

BoboTiG (approved):

BoboTiG (approved):

PR #2297, opened 2020-12-10
NXDRIVE-2309: Improve the session pause behavior

BoboTiG:
Before the patch, all items to be transferred were started and then paused when their session was paused. This was quite bad because as a transfer was initiated, a call to the server was made for at least retrieve a `batchId`. Even worse, all remaining transfers not yet handled were started _at once_; resulting in N parallel calls to the server and entries in the `Uploads` database table.

Now, when pausing a session, ongoing transfers will be finished for non-chunked ones, the others will be paused ASAP.

Another issue was fixed when starting the app and a session is paused: all its transfers were added to the queue and then filtered out by the `Processor`. The SQL query was improved to prevent pushing such pairs into the queue.

As always: less is better, let's be lazy :)

### DIFF
--- a/nxdrive/engine/dao/sqlite.py
+++ b/nxdrive/engine/dao/sqlite.py
@@ -1297,32 +1297,38 @@ class EngineDAO(ConfigurationDAO):
         )
 
     def _get_to_sync_condition(self) -> str:
-        return "pair_state != 'synchronized' AND pair_state != 'unsynchronized'"
+        return "pair_state NOT IN ('synchronized', 'unsynchronized')"
 
     def register_queue_manager(self, manager: "QueueManager") -> None:
-        # Prevent any update while init queue
+        """Register the queue manager and add all *pairs* to handle into the queue."""
+
         with self.lock:
             self.queue_manager = manager
-            con = self._get_write_connection()
-            c = con.cursor()
-            # Order by path to be sure to process parents before children
-            pairs: List[DocPair] = c.execute(
-                "SELECT *"
-                "  FROM States "
-                f"WHERE {self._get_to_sync_condition()}"
+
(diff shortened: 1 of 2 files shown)


nxdrive/engine/processor.py
```diff
@@ -549,7 +560,7 @@ def _direct_transfer_end(
         self.dao.remove_state(doc_pair, recursive=recursive)
 
         # Update session then handle the status
-        session = self.dao.get_session(doc_pair.session)
+        session = session or self.dao.get_session(doc_pair.session)
```
rom1win, line 563: I think we should always get the session from the database here as it can change many times before attaining `_direct_transfer_end()`

BoboTiG: Should we remove the `session` argument and always fetch a fresh session object?

nxdrive/engine/dao/sqlite.py
```diff
@@ -2778,19 +2784,19 @@ def resume_session(self, uid: int) -> None:
...
 
+            # Push not-yet-handled session transfers into the queue
             if self.queue_manager:
-                rows = c.execute(
-                    "SELECT * FROM States WHERE id IN"
-                    " (SELECT doc_pair FROM Uploads WHERE doc_pair IN (SELECT id FROM States WHERE Session = ?))",
-                    (uid,),
-                ).fetchall()
-                for doc_pair in rows:
+                rows = c.execute("SELECT * FROM States WHERE session = ?", (uid,))
```
BoboTiG, line 2796: Hm I just found something here: we potentially repush the same `doc_pair` for all ongoing uploads. And duplicates may happen. Right?

rom1win (approved):

PR #2414, opened 2021-01-26
NXDRIVE-2329: [Direct Transfer] Give more space to the QTreeView

BoboTiG:
Also:
- Reduce options height: they seem more "grouped" now.
- Reduce the new folder input width.


rom1win (approved):

BoboTiG:
Before: 
Capture d’écran 2021-01-26 à 13 47 03

After: 
Capture d’écran 2021-01-26 à 13 52 36

PR #2477, opened 2021-02-17
NXDRIVE-2524: [GNU/Linux] WAL journal mode cause manager database corruption

rom1win:
The DELETE journal mode is now user for the manager database instead of WAL.

### DIFF
--- a/nxdrive/engine/dao/sqlite.py
+++ b/nxdrive/engine/dao/sqlite.py
@@ -170,6 +170,7 @@ class AutoRetryConnection(Connection):
 class ConfigurationDAO(QObject):
 
     _state_factory: Type[Row] = DocPair
+    _journal_mode: str = "WAL"
 
     def __init__(self, db: Path, /) -> None:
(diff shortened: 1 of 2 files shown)


nxdrive/engine/dao/sqlite.py
```diff
@@ -443,6 +444,9 @@ def get_int(self, name: str, /, *, default: int = 0) -> int:
 class ManagerDAO(ConfigurationDAO):
 
     _state_factory = EngineDef
+    _journal_mode: str = (
+        "DELETE"  # WAL not needed as we write less often (NXDRIVE-2524).
+    )
```
BoboTiG, line 449:
```suggestion
    # WAL not needed as we write less often and it may have issues on GNU/Linux (NXDRIVE-2524)
    _journal_mode: str = "DELETE"
```

BoboTiG (commented):
Missing changelog entry.

BoboTiG (approved):

PR #2491, opened 2021-02-19
NXDRIVE-2523: Prevent crashes because of too many open file descriptors

BoboTiG:
The issue was not visible quickly but when Drive was running a long time with a lot of synchronization actions. Each thread was opening a connection to the SQLite database file, but it was never released.


BoboTiG:
Closing in favor of #2497, which is less intrusive.

PR #2522, opened 2021-03-04
NXDRIVE-2553: Adapt the code following recent Python client changes

BoboTiG:
- NXDRIVE-2552: Pass a custom callback for token renewal on S3


BoboTiG:
Requires `nuxeo` 5.0.0. It will be merged next after.

PR #2560, opened 2021-03-18
NXDRIVE-2558: Add a custom metric for hard crashes

rom1win:


nxdrive/commandline.py
```diff
@@ -739,6 +742,8 @@ def __enter__(self) -> "HealthCheck":
             if self.crash_file.is_file():
                 log.warning("It seems the application crashed at the previous run 😮")
                 State.has_crashed = True
+                with open(str(self.crash_file), "r") as my_file:
```
BoboTiG, line 745: You should leverage the `pathlib` power here:
```python
State.crash_details = self.crash_file.read_text().strip()
```
Also, to be safe, you should surround that new code with a try/except or `contextlib.suppress()` to prevent `OSError`s from starting the app.

nxdrive/gui/application.py
```diff
@@ -648,6 +652,17 @@ def _root_deleted(self) -> None:
             engine.reinit()
             engine.start()
 
+    def _send_crash_metrics(self) -> None:
+        for engine in self.manager.engines.copy().values():
```
BoboTiG, line 656: You should return early in case the app did not crash.

nxdrive/gui/application.py
```diff
@@ -648,6 +652,17 @@ def _root_deleted(self) -> None:
...
+            if engine.remote and State.has_crashed:
+                if not State.crash_details:
+                    engine.remote.metrics.send({CRASHED_HIT: 1})
```
BoboTiG, line 659: Move metrics into a dict. And update that dict with `State.crash_details` if available. That will make the code more readable and skip the need for a if/else block code.

nxdrive/gui/application.py
```diff
@@ -648,6 +652,17 @@ def _root_deleted(self) -> None:
...
+                else:
+                    engine.remote.metrics.send(
+                        {CRASHED_HIT: 1, CRASHED_TRACE: State.crash_details}
+                    )
+                return
```
BoboTiG, line 664: Simple `break` should be enough, right?

nxdrive/metrics/constants.py
```diff
@@ -36,4 +36,5 @@
 
 # Other
 CRASHED_HIT = "app.crashed.hit"
+CRASHED_TRACE = "app.crashed.trace"
```
BoboTiG, line 39: Changelog entry needed.

nxdrive/state.py
```diff
@@ -13,4 +13,4 @@
 """
 from types import SimpleNamespace
 
-State = SimpleNamespace(about_to_quit=False, has_crashed=False)
+State = SimpleNamespace(about_to_quit=False, has_crashed=False, crash_details="")
```
BoboTiG, line 16: Keep it sorted.

nxdrive/commandline.py
```diff
@@ -747,14 +752,28 @@ def __enter__(self) -> "HealthCheck":
 
         return self
 
-    def __exit__(self, *_: Any) -> None:
+    def __exit__(
```
BoboTiG, line 755: All changes here are not needed IMO. When a crash happens, that method will never be taken. The only way to record the stacktrace is in: https://github.com/nuxeo/nuxeo-drive/blob/6ff14d34c07b7edcc855b30fa3b957ebb7f352a6/nxdrive/fatal_error.py#L260-L265
You should use that and save `full_error` into the crash file.

nxdrive/commandline.py
```diff
@@ -577,12 +579,13 @@ def launch(self, options: Optional[Namespace], /, *, console: bool = False) -> i
                 log.warning(f"{APP_NAME} is already running: exiting.")
             return 0
 
+        exit_code: int = 1
```
BoboTiG, line 582: Does `mypy` throw a warning/error and you had to fix it like that?

nxdrive/gui/application.py
```diff
@@ -225,6 +226,9 @@ def __init__(self, manager: "Manager", *args: Any) -> None:
         # to self.aboutToQuit can run beforehand.
         self.aboutToQuit.connect(self.manager.stop)
 
+        # Send previous crash metrics
+        self._send_crash_metrics()
```
BoboTiG, line 230: Out of curiosity, why did you move that logic from `Manager` to `Application`?

BoboTiG (changes requested):

nxdrive/commandline.py
```diff
@@ -739,6 +741,8 @@ def __enter__(self) -> "HealthCheck":
             if self.crash_file.is_file():
                 log.warning("It seems the application crashed at the previous run 😮")
                 State.has_crashed = True
+                with suppress():
```
BoboTiG, line 744: Hm the exception type is missing there. Filter out `OSError`.

nxdrive/commandline.py
rom1win: Not really, my handling of the exception in the `HealthCheck.__exit__()` made it so that the code after `exit_code = app.exec_()` is still called in case of a crash. I had an error since in case of a crash, `exit_code` is undefined.

rom1win: And the `: int` was already there before

nxdrive/commandline.py
```diff
@@ -747,7 +751,7 @@ def __enter__(self) -> "HealthCheck":
 
         return self
 
-    def __exit__(self, *_: Any) -> None:
+    def __exit__(self, *_: Any) -> bool:
```
BoboTiG: Not needed.

nxdrive/fatal_error.py
```diff
@@ -262,9 +262,12 @@ def show_critical_error() -> None:
 
     import traceback
 
+    crash_file = Options.nxdrive_home / "crash.state"
     full_error = "".join(traceback.format_exception(*sys.exc_info()))
 
     try:
+        with suppress():
```
BoboTiG, line 269: Filter out `Exception`.

nxdrive/gui/application.py
```diff
@@ -648,6 +652,19 @@ def _root_deleted(self) -> None:
             engine.reinit()
             engine.start()
 
+    def _send_crash_metrics(self) -> None:
+        if not State.has_crashed:
+            return
+
+        metrics = {CRASHED_HIT: 1}
+        if State.crash_details:
+            metrics.update({CRASHED_TRACE: State.crash_details})
```
BoboTiG, line 661:
```suggestion
            metrics[CRASHED_TRACE] = State.crash_details
```

nxdrive/gui/application.py
rom1win: If my memory is correct, it was because in `Manager.load()` the remotes inside the engines are not yes instanciated.

nxdrive/fatal_error.py
```diff
@@ -262,9 +262,12 @@ def show_critical_error() -> None:
 
     import traceback
 
+    crash_file = Options.nxdrive_home / "crash.state"
     full_error = "".join(traceback.format_exception(*sys.exc_info()))
 
     try:
+        with suppress(Exception):
+            crash_file.write_text(full_error)
```
BoboTiG, line 270: Move those 3 lines ahead of the try/except block + blank line:
```python
crash_file = Options.nxdrive_home / "crash.state"
with suppress(Exception):
    crash_file.write_text(full_error)
```

BoboTiG (approved):

PR #2674, opened 2021-05-03
NXDRIVE-2612: Save all features state in the local configuration file

rom1win:


tests/functional/test_server_options.py
```diff
@@ -88,6 +90,13 @@ def test_features(feature, feat_name, default, manager_factory, tmp_path):
     updater = ServerOptionsUpdater(manager)
     opt_name = f"feature_{feat_name}"
 
+    paths, default_path = config_paths()
+    res = [conf_file for conf_file in paths if conf_file.is_file()]
+    conf_path = res[0] if res else default_path
```
BoboTiG, line 95: That whole logic is a copy-paste from the beginning of `save_config()`.
Shouldn't we refactor that part? Maybe by putting the logic into a new `config_path()` function:
```python
def config_path() -> Path:
    """TODO"""
    paths, default_path = config_paths()
    files = [file for file in paths if file.is_file()]
    return files[0] if files else default_path
```
And using it everywhere needed.

BoboTiG (approved):

BoboTiG (approved):

PR #2726, opened 2021-05-19
NXDRIVE-2217: Use the PRAGMA user_version to store the database version

rom1win:


nxdrive/engine/dao/sqlite.py
```diff
@@ -204,17 +204,19 @@ def __init__(self, db: Path, /) -> None:
         c = self.conn.cursor()
         self._init_db(c)
         if exists:
-            res = c.execute(
-                "SELECT value FROM Configuration WHERE name = ?", (SCHEMA_VERSION,)
-            ).fetchone()
+            res = c.execute("PRAGMA user_version").fetchone()
```
BoboTiG, line 207: Move all that logic into `get_user_version()`. I do not like the name, I would prefer `get_schema_version()` to handle it. And same for `set_schema_version()` (not `set_user_version()`, let's hide implementation details).

Here is the overall view:
```python
# 1) Rename get_schema_version() to current_schema_version() everywhere.
# 2) Then define new methods:

def get_schema_version(self, cursor, db_exists):
    version = int(cursor.execute("PRAGMA user_version").fetchone()[0])
    if version == 0 and db_exists:
        # Backward  compatibility
        version = SELECT value FROM Configuration WHERE name...
        self.set_schema_version(cursor, version)
        # and delete the Configuration entry
    return version
        
def set_schema_version(self, cursor, version):
    cursor.execute(f"PRAGMA user_version = {version}")
```

BoboTiG (approved):

PR #2754, opened 2021-05-26
NXDRIVE-2548: Databases refactoring

rom1win:


nxdrive/engine/dao/__init__.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import PosixPath, WindowsPath
+from sqlite3 import register_adapter
+
+from nxdrive.constants import WINDOWS
```
BoboTiG, line 4: Use relative import.

PR #2801, opened 2021-06-08
NXDRIVE-2677: Add custom metrics about the uploaded file

BoboTiG:


nxdrive/client/uploader/direct_transfer.py
```diff
@@ -96,6 +101,9 @@ def upload(
                 headers={
                     REQUEST_METRICS: json.dumps(
                         {
+                            DT_FILE_EXTENSION: file_path.suffix,
```
rom1win, line 104: What if there is no suffix ? Should we send somthing like "None" instead of an empty string ?
We could also remove the metrics if there is no suffix.

BoboTiG: I think it is interesting to keep the metric even with no suffix.
Maybe something like "NA" (Not Available) is clear enough, WDYT? Or even "n/a", it will be clear that there is no suffix.

rom1win (approved):

PR #2894, opened 2021-07-05
NXDRIVE-2229: [Direct Transfer] Use a proper NXQL query to fetch serv…

BoboTiG:
…er folders


BoboTiG:
It requires the next version of the `nuxeo` module.

PR #3277, opened 2021-11-25
NXDRIVE-2753: Create a diagram of the GUI part of the application

mr-shekhar:


rom1win (commented):
Is there a need to add the XML file ?
Make the pdf file visible from https://github.com/nuxeo/nuxeo-drive/blob/master/DEVELOPERS.md
Add a new section for the GUI inside the DEVELOPERS.md

PR #3513, opened 2022-06-29
NXDRIVE-2774: Make Document Type Selection Feature

swetayadav1:
Added Document Type Selection feature. Its state will depend on Direct Transfer feature state.
Folder Type and Document Type  dropdown state on Direct Transfer page will depend on Document Type Selection feature state.


swetayadav1:
Need to remove unwanted commits

PR #3659, opened 2022-11-03
NXDRIVE-1999: Direct Transfer let user select Document Type

swetayadav1:


swetayadav1:
Updated code available in different branch

PR #3749, opened 2023-01-17
NXDRIVE-2801: Do not prefix synchronized folder name with ancestors when there is no duplicated root

gitofanindya:
NXDRIVE-2801: Do not prefix synchronized folder name with ancestors when there is no duplicated root


mr-shekhar (approved):
Pl. check, for multi-lingual.
Rest 👍

swetayadav1 (approved):

PR #3820, opened 2023-03-02
NXDRIVE-2801: Do not prefix synchronized folder name with ancestors when there is no duplicated root

gitofanindya:


nxdrive/engine/processor.py
```diff
@@ -1474,6 +1478,13 @@ def _create_remotely(
                     f"in {self.local.abspath(local_parent_path)!r}"
                 )
                 return self.local.make_folder(local_parent_path, name)
+            elif doc_pair.folderish:
+
+                log.info(
```
mr-shekhar, line 1483: Code is same in both the conditions

nxdrive/engine/processor.py
```diff
@@ -1474,6 +1478,13 @@ def _create_remotely(
...
+                    f"Creating local folder {name!r} "
+                    f"in {self.local.abspath(local_parent_path)!r}"
+                )
+                return self.local.make_folder(local_parent_path, name)
```
mr-shekhar, line 1487: same return st.in both the conditions.

tests/functional/test_create_sync_folder.py
```diff
@@ -10,12 +10,14 @@ def test_create_folder(manager_factory, tmp_path):
...
+    dp.folderish = True
     create_rem = test_processor._create_remotely
+
     full_path = create_rem(dp, dp, "default - domain - Workspaces - Administrator")
     assert str(path.join(tmp_path, "Administrator")) == str(full_path)
-    full_path_duplicated = create_rem(
-        dp, dp, "default - domain1 - Workspaces - Administrator"
-    )
+
+    full_path = create_rem(dp, dp, "default - domain1 - Workspaces - Administrator")
```
mr-shekhar, line 20: What is the change here?

gitofanindya: we are using same variable to check multiple cases. because after first assert we do not need the variable again. it will take less memory.

nxdrive/engine/processor.py
gitofanindya: done.

gitofanindya: refactored the code.

swetayadav1 (approved):

mr-shekhar (approved):
👍

PR #3969, opened 2023-05-23
NXDRIVE-2838: Fix Release build

poojadaine:
NXDRIVE-2838: Fix Release builld

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -202,6 +202,7 @@ jobs:
       - name: Setup git
         run: |
           git config user.email ${{ env.GITHUB_EMAILID }}
+          git config user.name ${{ env.GITHUB_USERNAME }}
           git fetch --unshallow --tags
           git remote set-url origin "https://${{ env.GITHUB_USERNAME }}:${{ secrets.NXDRIVE_GIT_TOKEN }}@github.com/nuxeo/nuxeo-drive.git"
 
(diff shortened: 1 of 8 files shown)


tools/release.sh
```diff
@@ -28,7 +28,8 @@ create() {
 
     echo ">>> [${release_type} ${drive_version}] Creating the tag"
     git tag -f -a "${release_type}-${drive_version}" -m "Release ${drive_version}"
-    git push -f origin "${release_type}-${drive_version}"
+    #git push -f origin "${release_type}-${drive_version}"
```
mr-shekhar, line 31: is this req.?

.github/workflows/release.yml
```diff
@@ -199,8 +198,8 @@ jobs:
 
       - name: Setup git
         run: |
-          git config user.email "maintainers-python+github@nuxeo.com"
-          git config user.name "GitHub-CI action (Release)"
+          git config user.email "drive@hyland.com"
```
mr-shekhar, line 201: use these as var, as they are used in several places.

mr-shekhar (changes requested):
Also update other required files

mr-shekhar (changes requested):
Also update other required files

tools/release.sh
swetayadav1: Updated code

.github/workflows/release.yml
swetayadav1: Done

mr-shekhar (approved):
👍

gitofanindya (approved):

PR #4876, opened 2024-05-15
NXDRIVE-2926: Update github Action Runner to use mac-latest

gitofanindya:
NXDRIVE-2926: Update github Action Runner to use mac-latest


mr-shekhar (approved):
👍

PR #5852, opened 2025-06-06
NXDRIVE-3000 : Fix code scanning issues

S-Ghosh-Dev2:
## Summary by Sourcery

CI:
- Set contents: read and pull-requests: write permissions across all GitHub Actions workflows


gitofanindya (approved):

.github/workflows/alpha_cleanup.yml
```diff
@@ -1,5 +1,8 @@
 # Alpha purgation
 name: Alpha clean-up
+permissions:
+  contents: read
+  pull-requests: write
```
mr-shekhar, line 5: Why do we need provide write requests to pull_request

S-Ghosh-Dev2: From what I understand, we need to have write access to open a pull request and change the code directly in a public repository. If we do not provide write access, then we cannot make changes to that repository, we would need to fork that repository. 

Please see the link : 
[https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request](url)

PR #6357, opened 2026-04-24
NXDRIVE-3163: Align on user and group generated UUID at Drive

gitofanindya:
## Summary by Sourcery

Persist and restore the server-side user UUID to keep the userid mapper consistent across sessions and use it to resolve users in direct edit flows.

New Features:
- Store the server-side user UUID in the local configuration when binding or updating credentials.
- Restore the userid mapper from the persisted user UUID at engine initialization, prompting re-login if the UUID is missing.
- Resolve engines and user display names via the userid mapper so UUIDs can be mapped back to usernames in direct edit operations and error dialogs.

Bug Fixes:
- Avoid calling remote token update when the remote endpoint is not yet initialized, and ensure token encryption uses the up-to-date username.


S-Ghosh-Dev2 (approved):
+1

